PR #6, opened 2024-11-21
feat: implement Speaker and Serdes

### LINKED ISSUE #3: Implement the CoderVPN protocol
We use a simple control protocol, called the CoderVPN protocol for communication between the user app and the Windows service, and between the Windows service and the `coder daemon` process.

Image

This task involves implementing the protocol framework in C#

This protocol uses Protobufs for message serialization, but does *not* use dRPC like our other Protobuf-based APIs since dRPC is not available for the languages we’d like to use for native OS development (e.g. C#, Swift).

The protocol runs over a single bidirectional stream between the native code (called the Manager) and the Go code (called the Tunnel). We keep everything simple and send messages serially on the stream (rather than rely on a multiplexing library like yamux, which is only implemented in Go). This means that RPC responses might get behind other, unrelated messages, but we expect this control protocol to be very light in traffic, and response times are generally not crucial.

On opening the stream, each side sends a single plaintext header of the form `codervpn <major.minor> <role>\n` where \<major.minor\> is the protocol version and \<role\> is “manager” or “tunnel”. After reading and verifying the other side’s header, they each proceed to send a series of messages encoded as:

- 4-byte unsigned integer, big-endian: length
- \<length\> bytes of Protobuf encoded message, either `TunnelMessage` or `ManagerMessage` according to their role, as follows:

```protobuf
// RPC allows a very simple unary request/response RPC mechanism.  The requester
// generates a unique msg_id which it sets on the request, the responder sets
// response_to that msg_id on the response message
message RPC {
    uint64 msg_id = 1;
    uint64 response_to = 2;
}

// ManagerMessage is a message from the manager (to the tunnel).
message ManagerMessage {
    RPC rpc = 1;
    oneof msg {
        GetPeerUpdate get_peer_update = 2;
        NetworkSettingsResponse network_settings = 3;
        StartRequest start = 4;
        StopRequest stop = 5;
    }
}

// TunnelMessage is a message from the tunnel (to the manager).
message TunnelMessage {
    RPC rpc = 1;
    oneof msg {
        Log log = 2;
        PeerUpdate peer_update = 3;
        NetworkSettingsRequest network_settings = 4;
        StartResponse start = 5;
        StopResponse stop = 6;
    }
}

// Log is a log message generated by the tunnel.  The manager should log it to
// the system log. It is one-way tunnel -> manager with no response.
message Log {
    enum Level {
        UNSPECIFIED = 0;
        INFO = 1;
        WARN = 2;
        ERROR = 3;
        CRITICAL = 4;
        FATAL = 5;
    }
    Level level = 1;

    string message = 2;
    repeated string logger_names = 3;

    message Field {
        string name = 1;
        string value = 2;
    }
    repeated Field fields = 4;
}

// GetPeerUpdate asks for a PeerUpdate with a full set of data.
message GetPeerUpdate {}

// PeerUpdate is an update about workspaces and agents connected via the tunnel.
// It is generated in response to GetPeerUpdate (which dumps the full set). It is
// also generated on any changes (not in response to any request).
message PeerUpdate {
    repeated Workspace upserted_workspaces = 1;
    repeated Agent upserted_agents = 2;
    repeated Workspace deleted_workspaces = 3;
    repeated Agent deleted_agents = 4;
}

message Workspace {
    bytes id = 1; // UUID
    string name = 2;
    
  enum Status {
		UNKNOWN = 0;
		PENDING = 1;
		STARTING = 2;
		RUNNING = 3;
		STOPPING = 4;
		STOPPED = 5;
		FAILED = 6;
		CANCELING = 7;
		CANCELED = 8;
		DELETING = 9;
		DELETED = 10;
	}
	Status status = 3;
}

message Agent {
    bytes id = 1; // UUID
    string name = 2;
    bytes workspace_id = 3; // UUID
    string fqdn = 4;
    repeated string ip_addrs = 5;
    // last_handshake is the primary indicator of whether we are connected to a
    // peer. Zero value or anything longer than 5 minutes ago means there is a
    // problem.
    google.protobuf.Timestamp last_handshake = 6;
}

// NetworkSettingsRequest is based on
// https://developer.apple.com/documentation/networkextension/nepackettunnelnetworksettings
// for macOS.  It is a request/response message with response
// NetworkSettingsResponse
message NetworkSettingsRequest {
    uint32 tunnel_overhead_bytes = 1;
    uint32 mtu = 2;

    message DNSSettings {
        repeated string servers = 1;
        repeated string search_domains = 2;
        // domain_name is the primary domain name of the tunnel
        string domain_name = 3;
        repeated string match_domains = 4;
        // match_domains_no_search specifies if the domains in the matchDomains
        // list should not be appended to the resolver’s list of search domains.
        bool match_domains_no_search = 5;
    }
    DNSSettings dns_settings = 3;

    string tunnel_remote_address = 4;

    message IPv4Settings {
        repeated string addrs = 1;
        repeated string subnet_masks = 2;
        // router is the next-hop router in dotted-decimal format
        string router = 3;

        message IPv4Route {
            string destination = 1;
            string mask = 2;
            // router is the next-hop router in dotted-decimal format
            string router = 3;
        }
        repeated IPv4Route included_routes = 4;
        repeated IPv4Route excluded_routes = 5;
    }
    IPv4Settings ipv4_settings = 5;

    message IPv6Settings {
        repeated string addrs = 1;
        repeated uint32 prefix_lengths = 2;

        message IPv6Route {
            string destination = 1;
            uint32 prefix_length = 2;
            // router is the address of the next-hop
            string router = 3;
        }
        repeated IPv6Route included_routes = 3;
        repeated IPv6Route excluded_routes = 4;
    }
    IPv6Settings ipv6_settings = 6;
}

// NetworkSettingsResponse is the response from the manager to the tunnel for a
// NetworkSettingsRequest
message NetworkSettingsResponse {
    bool success = 1;
    string error_message = 2;
}

// StartRequest is a request from the manager to start the tunnel.  The tunnel
// replies with a StartResponse.
message StartRequest {
    int32 tunnel_file_descriptor = 1;
    string coder_url = 2;
    string api_token = 3;
}

message StartResponse {
    bool success = 1;
    string error_message = 2;
}

// StopRequest is a request from the manager to stop the tunnel. The tunnel
// replies with a StopResponse.
message StopRequest {}

// StopResponse is a response to stopping the tunnel. After sending this response,
// the tunnel closes its side of the bidirectional stream for writing.
message StopResponse {
    bool success = 1;
    string error_message = 2;
}
```

deansheather:
Closes #3

- Adds protobuf code generation
- Adds utilities and wrapping code around RPC message types
- Adds RPC protocol code required for sending and receiving messages
- Adds a end-to-end test with messages and replies between two Speakers

## TODO:
- [x] Add the `csharp_namespace` option to the upstream `vpn.proto`
- [x] Decide between event delegates (callbacks) or channels for receiving messages
- [x] Plumb `CancellationTokenSource` into more operations (any async `Speaker` operation should use the supplied `CancellationToken` as well as the global `CancellationTokenSource`)
- [x] Make the startup process async?
- [x] Try to declutter the generic type specifications around the codebase
    - It may be possible to condense the two `TS/TSi` and `TR/TRi` into just `TS` and `TR` with some finangling and changes around the `RpcMessage<T>` type
        - Maybe switch (back) to an interface rather than an abstract class
- [x] Move the role to be attached to the type so it doesn't need to be passed in manually to the `Speaker` at all
- [x] Tests
    - [x] ApiVersion
    - [x] RpcRole
    - [x] RpcHeader
    - [x] RpcMessage (and RpcRoleAttribute)
    - [x] Serdes
    - [x] Speaker


Rpc/Serdes.cs
```diff
@@ -0,0 +1,97 @@
...
+        if (len > MaxMessageSize)
+            throw new IOException($"Received message size {len} exceeds maximum {MaxMessageSize}");
+
+        var msgBytes = new byte[len];
+        await conn.ReadExactlyAsync(msgBytes, ct);
+
+        try
+        {
+            var msg = _parser.ParseFrom(msgBytes);
+            if (msg?.RpcField is null)
```
ethanndickson, line 88: Messages from the tunnel that aren't a response and don't require a response have a null `RPC` field. My understanding of the design is that the presence of `MsgID` indicates whether a response is expected.

deansheather: @spikecurtis thoughts on changing this behavior in the protocol? I think all messages should have RPC set and have a message ID. If you want to enforce on each end that replies can only be sent to messages that expect one, we could add a `ExpectReply bool` to RPC or something perhaps

deansheather: It's like having a database table where the id field is optional... if not all messages have an ID then it should be called something else, like `reply_with_id`

During debugging it might be helpful to also log sent and received messages, and it's nice if every message has an ID that can be logged on both ends of the connection

Rpc.Proto/vpn.proto
ethanndickson, line 1: Remember to update this file with my changes on `coder/coder`

Rpc/Utilities/TaskUtilities.cs
```diff
@@ -0,0 +1,59 @@
...
+    ///         await TaskUtilities.CancellableWhenAll(cts, task1, task2);
+    ///     </code>
+    /// </example>
+    /// <param name="tasks">Tasks to wait on</param>
+    /// <param name="cts">The cancellation token source that was provided to each task</param>
+    /// <returns>
+    ///     A task that completes when all tasks are completed, with the cancellation or exception state of the first
+    ///     non-successful task
+    /// </returns>
+    public static async Task CancellableWhenAll(CancellationTokenSource cts, params Task[] tasks)
```
ethanndickson, line 25: nice!

ethanndickson (approved):
I'm happy with the repo structure, the namespaces, and I of course agree that the C# code should be idiomatic C# and not just a translation of the Go code. FWICT the C# API looks good.

LGTM, aside from the earlier `RPC.MsgID` thread - can probably just merge this for now?

Rpc/Speaker.cs
```diff
@@ -0,0 +1,246 @@
...
+                var message = await _serdes.ReadMessage(_conn, ct);
+                if (message.RpcField.ResponseTo != 0)
+                    // Look up the TaskCompletionSource for the message ID and
+                    // complete it with the message.
+                    if (_pendingReplies.TryRemove(message.RpcField.ResponseTo, out var tcs))
+                        tcs.SetResult(message);
+
+                // TODO: we should log unknown replies
+                // Start a new task in the background to handle the message.
+                _ = Task.Run(() => Receive?.Invoke(new ReplyableRpcMessage<TS, TR>(this, message)), ct);
```
spikecurtis, line 167: If I'm reading this correctly, if we get a reply message we _both_ return it to the requestor and trigger the receive event. Is that really what we want? The event handler would have to check again to make sure we're not also treating it as an unsolicited message.

Rpc/Speaker.cs
```diff
@@ -0,0 +1,246 @@
...
+    ///     Send a message without waiting for a reply. If a reply is received it will be handled by the callback.
+    /// </summary>
+    /// <param name="message">Message to send</param>
+    /// <param name="ct">Optional cancellation token</param>
+    public async Task SendMessage(TS message, CancellationToken ct = default)
+    {
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(ct, _cts.Token);
+        message.RpcField = new RPC
+        {
+            MsgId = Interlocked.Add(ref _lastMessageId, 1),
```
spikecurtis, line 190: As discussed on our call, we need to distinguish between messages that expect a reply and those that don't in terms of something on the wire.  The original design had the RPC omitted for messages that don't expect a reply.

spikecurtis (approved):
Some inline comments, but overall, I like the shape of the APIs to the protocol.

Rpc/Speaker.cs
deansheather: Yeah this was a mistake, I've added a `continue` to avoid writing replies to the event handler

Rpc/Serdes.cs
deansheather: I've updated the behavior to match Go for now and I talked with Spike today about making some changes to the name of the MsgId field which I'll do later

PR #13, opened 2025-01-28
chore: add GitHub actions and remove Package wapproj

### LINKED ISSUE #7: Add GitHub Actions automation
- Linting
- Formatting
- Testing

deansheather:
Adds fmt, test and build jobs using dotnet.

The fmt job currently doesn't match what Rider does, but I think that will be difficult to pull off without using a different formatting tool which I don't want to investigate for now.

Removes the `Package` project in favor of making the `App` project include the necessary stuff for packaging itself. We weren't getting any benefit from it being in a separate package yet and it was not building correctly in dotnet (only in VS) when they were separated.

Closes #7


deansheather:
During the actual publish when we get release CI going, we probably want to make our build matrix manually specify `RuntimeIdentifier` and `PublishReadyToRun` on the publish command.

https://github.com/dotnet/sdk/issues/39141

App/Package.appxmanifest
```diff
@@ -0,0 +1,52 @@
...
+<Package
+    xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest"
+    xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+    xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+    xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+    IgnorableNamespaces="uap rescap">
+
+    <Identity
+        Name="925b49fc-4648-4967-b4e6-b5473061ee62"
+        Publisher="CN=dean"
```
spikecurtis, line 12: Added via Giphy

spikecurtis (approved):

App/Package.appxmanifest
deansheather: lmfao oops, I guess it defaults to the current user

PR #14, opened 2025-01-28
fix: various fixes for tunnel startup to work

### LINKED ISSUE #2: Build a Windows Service that implements CoderVPN
Create a native Windows Service that auto-starts with Windows and creates a named pipe to accept connections from the user application.

Image

On connection, it uses the CoderVPN protocol to communicate with the user application.  When the user application starts the VPN, it uses the provided credentials to connect to the Coder deployment, download the Coder binary, create a pair of communication pipes, then fork/exec to start the `coder daemon` process (coder/coder#14735).

deansheather:
Fixes various problems that prevented tunnel startup from succeeding and prevented clients from connecting to the RPC server.

Added a new temporary package Vpn.DebugClient to help control the system service component until the UI is ready.

Adds utility PS1 scripts for adding the debug binary to a new system service and managing/hotswapping it.
 
Start and stop operations from a client now reliably work, although there are various TODOs around the service codebase still.

Closes #2


Vpn.Service/Manager.cs
```diff
@@ -79,24 +80,46 @@ private async Task HandleClientMessageStart(ReplyableRpcMessage<ServiceMessage,
...
+                    Start = new StartResponse
+                    {
+                        Success = true,
+                    },
+                }, ct);
+                return;
+            }
 
+            _lastStartRequest = message.Message.Start;
+            _lastServerVersion = serverVersion;
```
spikecurtis, line 103: It doesn't seem like these are protected from concurrent access.

Vpn.Service/Manager.cs
```diff
@@ -79,24 +80,46 @@ private async Task HandleClientMessageStart(ReplyableRpcMessage<ServiceMessage,
...
+
+            // Stop the tunnel if it's running so we don't have to worry about
+            // permissions issues when replacing the binary.
+            await _tunnelSupervisor.StopAsync(ct);
             await DownloadTunnelBinaryAsync(message.Message.Start.CoderUrl, serverVersion, ct);
-            await _tunnelSupervisor.StartAsync(_config.TunnelBinaryPath, HandleTunnelRpcMessage,
-                HandleTunnelRpcError,
+            await _tunnelSupervisor.StartAsync(_config.TunnelBinaryPath, HandleTunnelRpcMessage, HandleTunnelRpcError,
```
spikecurtis, line 109: same with this section: we need to ensure that stops and starts are serialized so that you don't get `stop, stop, start, start` if two clients connect around the same time.

Vpn.Service/ManagerRpcService.cs
```diff
@@ -112,17 +135,28 @@ private async Task HandleRpcClientAsync(NamedPipeServerStream pipeServer, Cancel
...
     private async Task HandleRpcMessageAsync(ReplyableRpcMessage<ServiceMessage, ClientMessage> message,
         CancellationToken ct)
     {
         _logger.LogInformation("Received RPC message: {Message}", message.Message);
         await _manager.HandleClientRpcMessage(message, ct);
     }
+
+    public async Task BroadcastAsync(ServiceMessage message, CancellationToken ct)
+    {
+        foreach (var (clientId, client) in _activeClients)
```
spikecurtis, line 147: Is there anything that prevents `_activeClients` from being modified while you are looping over it?

deansheather: Since it's a concurrent dictionary it's exception-safe. As far as I understand, any added or removed items from the dictionary while we're looping may or may not get called, but I think that's fine.

Vpn.Service/Manager.cs
deansheather: Added a semaphore that fixes both of these issues

spikecurtis (approved):

PR #20, opened 2025-02-07
feat: Sign in Window views & view models

spikecurtis:
Adds views for the Sign In dialog window, and a backing view model.

This is still a bit rough around the edges, but I'd like to get some eyes on it and merge to iterate.

It also needs to be integrated with some of the structural changes in #19 including dependency injection. That can happen per or post merge, depending on when #19 lands in `main`.

Known issues:
 * Input validation. If you give an invalid URL, then clicking "Next" will crash the app since the hyperlink button won't accept an invalid URI.
 * Relatively janky Error message handling -- doesn't use any consistent styling.


spikecurtis:
* **#20** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/20?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/20?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

App/SignInTokenPage.xaml
```diff
@@ -0,0 +1,93 @@
...
+                Content="Generate a token via the Web UI"
+                NavigateUri="{x:Bind ViewModel.GenTokenURL, Mode=OneTime}" />
+
+        </Grid>
+
+        <StackPanel
+            Orientation="Horizontal"
+            HorizontalAlignment="Center"
+            Spacing="10">
+            <Button Content="Back" HorizontalAlignment="Right"
```
deansheather, line 79: I think the better way of doing forwards/backwards navigation is to use a `<RootFrame>` in the window rather than setting `Window.Content` directly.

One pitfall that might make it hard to use the RootFrame's built-in navigation stuff is that I don't think it lets you instantiate the Window types yourself - which makes it hard to get a ViewModel in there.

In the TrayWindow code in my PR, I ended up just setting `RootFrame.Content` directly because of this limitation (and because I don't need forwards/backwards navigation). But if you could get that working it'd be more idiomatic I think

App/TrayWindow.xaml
```diff
@@ -159,9 +159,10 @@
         <HyperlinkButton
             Margin="-10,0,-10,0"
             HorizontalAlignment="Stretch"
-            HorizontalContentAlignment="Left">
+            HorizontalContentAlignment="Left"
+            Click="{x:Bind SignIn_Click}">
 
-            <TextBlock Text="Sign out" Foreground="{ThemeResource DefaultTextForegroundThemeBrush}" />
+            <TextBlock Text="Sign in" Foreground="{ThemeResource DefaultTextForegroundThemeBrush}" />
```
deansheather, line 165: I think this will need to be reverted before merge

App/ViewModels/SignInViewModel.cs
```diff
@@ -0,0 +1,45 @@
...
+public partial class SignInViewModel : ObservableObject
+{
+    public SignInViewModel()
+    {
+        _url = string.Empty;
+        CoderToken = string.Empty;
+    }
+
+    [ObservableProperty]
+    private string _url;
```
deansheather, line 23: Try this instead to avoid the warnings:
```
[ObservableProperty]
public partial Type Name { get; set; } = DefaultValue;
```

If you want the setter to be private, use `private set;`. And since they're now public they should have capitalized named.

App/ViewModels/SignInViewModel.cs
```diff
@@ -0,0 +1,45 @@
...
+        // TODO: this should call into the backing model to do the login with _url and Token.
+        LoginError = "This is a placeholder error.";
+    }
+
+    public string GenTokenURL
+    {
+        get
+        {
+            // TODO: use a real URL parsing library
+            return _url + "/cli-auth";
```
deansheather, line 42: Try something like:
```cs
var baseUri = new Uri(_url);
var cliAuthUri = new Uri(baseUri, "/cli-auth");
return cliAuthUri.ToString();
```

App/SignInWindow.xaml.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    private const double WIDTH = 600.0;
+    private const double HEIGHT = 300.0;
+
+    private readonly SignInViewModel viewModel;
+    public SignInWindow(SignInViewModel vm)
+    {
+        viewModel = vm;
+        InitializeComponent();
+        var urlPage = new SignInURLPage(this, viewModel);
```
deansheather, line 34: Could you just instantiate both of these pages once and store them as private properties on the Window?

PR #33, opened 2025-02-28
chore: installer CI

deansheather:


.github/workflows/release.yaml
```diff
@@ -4,6 +4,12 @@ on:
   push:
     tags:
       - '*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version number (e.g. 1.2.3)'
+        required: true
+        default: '1.2.3'
```
matifali, line 12: Is this for testing only? Or do we wish to cut manual releases?

deansheather: yes, we will be pushing tags normally

sreya (approved):

PR #34, opened 2025-02-28
chore: change stale agent colour to yellow

ethanndickson:
On the Mac app, we use Yellow when an agent's last handshake was longer than 5 minutes ago. We should reserve red for a more fatal error, rather than something that will show on every connection.

### DIFF
--- a/App/ViewModels/AgentViewModel.cs
+++ b/App/ViewModels/AgentViewModel.cs
@@ -10,6 +10,7 @@ public enum AgentConnectionStatus
 {
     Green,
     Red,
+    Yellow,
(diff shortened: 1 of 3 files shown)


App/ViewModels/AgentViewModel.cs
```diff
@@ -10,6 +10,7 @@ public enum AgentConnectionStatus
 {
     Green,
     Red,
+    Yellow,
     Gray,
```
deansheather, line 14: Wouldn't yellow be above red in terms of severity?

deansheather (approved):

Issue #40, opened 2025-03-04
Better error notifications


deansheather:
Currently we don't have a good system for notifying the user when errors happen.
 
- Every error should result in a Windows notification.
- Every error when starting/stopping the VPN should show red text on the tray window
- Every login error should show red text in the login window

The stack trace (if applicable) should be easily reachable but shouldn't clog up the screen.

Relates to #22

PR #44, opened 2025-03-06
feat: show vpn start/stop failure in app

deansheather:
Adds red text that appears when the VPN fails to start or stop. After an error, any manual start/stop operation will clear the error.

Changes a lot of the optional components (`TextBlock`s, etc.) in the tray window to use bools for visibility rather than custom converters, because the logic behind whether to show them is slightly more complicated now. I could've maybe rearranged the app and shifted everything around into multiple parent containers for each condition that we check but it seemed harder than just using bools on the view model.

Most of the lines in this PR were from restructuring the XAML to make it simpler (and unindenting many lines in the process).

image

Contributes to #40


App/ViewModels/TrayWindowViewModel.cs
```diff
@@ -23,22 +25,46 @@ public partial class TrayWindowViewModel : ObservableObject
 
     private DispatcherQueue? _dispatcherQueue;
 
-    [ObservableProperty] public partial VpnLifecycle VpnLifecycle { get; set; } = VpnLifecycle.Unknown;
+    [ObservableProperty]
+    [NotifyPropertyChangedFor(nameof(ShowEnableSection))]
+    [NotifyPropertyChangedFor(nameof(ShowWorkspacesHeader))]
+    [NotifyPropertyChangedFor(nameof(ShowNoAgentsSection))]
+    [NotifyPropertyChangedFor(nameof(ShowAgentsSection))]
+    [NotifyPropertyChangedFor(nameof(ShowFailedSection))]
```
spikecurtis, line 33: ShowFailedSection doesn't depend on the lifecycle.

It really is annoying that we have to maintain these deps this way, where you have to mark the dependency, not the dependent. Harder to verify it's correct.

spikecurtis (approved):

App/ViewModels/TrayWindowViewModel.cs
deansheather: Yeah I definitely don't like it either. Since it's a decorator attribute I wonder if someone could write an inverse version of this using codegen or something. Maybe one already exists and I've just been doing it the dumb way

deansheather, line 33:
```suggestion
```

deansheather: Thanks for spotting this though!

PR #54, opened 2025-03-11
chore: add SizedFrame for window sizing

deansheather:
First part of sign-in revamp.

Adds a `SizedFrame` to unify some of the quirky window sizing code, as it's now used in `SignInWindow` as well as `TrayWindow`.

It'll also be used in `FileSyncListWindow`

### DIFF
--- a/App/Controls/SizedFrame.cs
+++ b/App/Controls/SizedFrame.cs
@@ -11,8 +11,11 @@ public class SizedFrameEventArgs : EventArgs
 }
 
 /// <summary>
-///     SizedFrame extends Frame by adding a SizeChanged event. Sadly this is necessary because
-///     Window.Content.SizeChanged doesn't trigger when the Page's content changes.
+///     SizedFrame extends Frame by adding a SizeChanged event, which will be triggered when:
+///       - The contained Page's content's size changes
+///       - We switch to a different page.
+///
+///     Sadly this is necessary because Window.Content.SizeChanged doesn't trigger when the Page's content changes.
 /// </summary>
 public class SizedFrame : Frame
 {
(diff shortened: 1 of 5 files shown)


App/Controls/SizedFrame.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Coder.Desktop.App.Controls;
+
+public class SizedFrameEventArgs : EventArgs
+{
+    public Size NewSize { get; init; }
+}
+
+/// <summary>
+///     SizedFrame extends Frame by adding a SizeChanged event. Sadly this is necessary because
```
spikecurtis, line 14: SizedFrame extends Frame by adding a SizeChanged event, and wiring up the event to trigger whenever:

 * The contained Page changes size
 * We switch to a different page

Sadly, this is necessary...

App/Views/TrayWindow.xaml.cs
```diff
@@ -120,55 +121,31 @@ public void SetRootFrame(Page page)
...
-        ResizeWindow();
-        MoveWindow();
+        RootFrame.SetPage(page);
     }
 
-    private void Content_SizeChanged(object sender, SizeChangedEventArgs e)
+    private void RootFrame_SizeChanged(object sender, SizedFrameEventArgs e)
     {
-        ResizeWindow();
+        ResizeWindow(e.NewSize.Width);
```
spikecurtis, line 129:
```suggestion
        ResizeWindow(e.NewSize.Height);
```

spikecurtis (approved):
LGTM, but need to fix Width vs Height (comment inline). I don't need to review again.

App/Views/TrayWindow.xaml.cs
deansheather: 🤦

PR #60, opened 2025-03-19
feat: add mock UI for file syncing listing

deansheather:
Adds a file sync listing window with mocked data. Since it's difficult (I think) to disable sections of XAML based on release/debug builds, this mocked window replaces the `Sign out` button on debug builds.

- Adds `FileSyncListWindow`, `FileSyncListViewModel`, `FileSyncListMainPage`
- Adds `MutagenSessionModel` as well as useful helpers
- Adds `DependencyObjectSelector` for mapping enum values in XAML
    - This is necessary as there doesn't seem to be another way of binding to a `ThemeResource` with a map

image
image


App/ViewModels/FileSyncListViewModel.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        ClearNewForm();
+        CreatingNewSession = true;
+    }
+
+    public async Task OpenLocalPathSelectDialog(Window window)
+    {
+        var picker = new FolderPicker
+        {
+            SuggestedStartLocation = PickerLocationId.ComputerFolder,
```
spikecurtis, line 151: Home directory seems more useful IMO

App/ViewModels/FileSyncListViewModel.cs
```diff
@@ -0,0 +1,188 @@
...
+using WinRT.Interop;
+
+namespace Coder.Desktop.App.ViewModels;
+
+public partial class FileSyncListViewModel : ObservableObject
+{
+    public delegate void OnFileSyncListStaleDelegate();
+
+    // Triggered when the window should be closed.
+    public event OnFileSyncListStaleDelegate? OnFileSyncListStale;
```
spikecurtis, line 20: I think it's confusing UX to just close out the window if something is wrong. It'll seem like a bug if you're creating a new sync session and the window just disappears.

If you end up creating a sync session that doesn't work for some reason, we should just show it in an error state.

App/Views/FileSyncListWindow.xaml
```diff
@@ -0,0 +1,20 @@
...
+
+<winuiex:WindowEx
+    x:Class="Coder.Desktop.App.Views.FileSyncListWindow"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:winuiex="using:WinUIEx"
+    mc:Ignorable="d"
+    Title="Coder Desktop"
```
spikecurtis, line 11:
```suggestion
    Title="Coder File Sync"
```

App/Views/FileSyncListWindow.xaml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<winuiex:WindowEx
```
spikecurtis, line 3: Not for this PR, but should we consider converting our other windows to WinUIEx?

App/Views/Pages/TrayWindowMainPage.xaml
```diff
@@ -118,22 +115,50 @@
...
+                                                <SolidColorBrush Color="#ffcc01" />
+                                            </converters:StringToBrushSelectorItem.Value>
+                                        </converters:StringToBrushSelectorItem>
+                                        <converters:StringToBrushSelectorItem Key="Green">
+                                            <converters:StringToBrushSelectorItem.Value>
+                                                <SolidColorBrush Color="#34c759" />
+                                            </converters:StringToBrushSelectorItem.Value>
+                                        </converters:StringToBrushSelectorItem>
+                                    </converters:StringToBrushSelector>
+                                </StackPanel.Resources>
```
spikecurtis, line 144: I find this way less readable than the `AgentStatusToColorConverter`

App/Converters/DependencyObjectSelector.cs
```diff
@@ -0,0 +1,155 @@
...
+using Microsoft.UI.Xaml.Data;
+using Microsoft.UI.Xaml.Media;
+
+namespace Coder.Desktop.App.Converters;
+
+// This file uses manual DependencyProperty properties rather than
+// DependencyPropertyGenerator since it doesn't seem to work properly with
+// generics.
+
+public class DependencyObjectSelectorItem<TK, TV> : DependencyObject
```
spikecurtis, line 15: What is this used for?

App/Models/MutagenSessionModel.cs
```diff
@@ -0,0 +1,310 @@
...
+
+    public MutagenSessionModel(State state)
+    {
+        Identifier = state.Session.Identifier;
+        Name = state.Session.Name;
+
+        // If the protocol isn't what we expect for alpha or beta, show
+        // "unknown".
+        if (state.Session.Alpha.Protocol == Protocol.Local && !string.IsNullOrWhiteSpace(state.Session.Alpha.Path))
+            LocalPath = state.Session.Alpha.Path;
```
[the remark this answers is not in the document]
ethanndickson: This would also be useful if we ever wanted to support file sync between two workspaces.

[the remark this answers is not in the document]
deansheather: If we implemented syncing between two workspaces it would likely be via a daemon running on one of the workspaces rather than synchronization going through your local computer.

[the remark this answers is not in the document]
ethanndickson: I don't really know what it'd look like in practice, but from a user's perspective they'd want to configure it via Coder Desktop, even if the sync persists with the app closed. In that case we'd need to get rid of the explicit Local & Remote distinction.

App/Views/Pages/FileSyncListMainPage.xaml
```diff
@@ -0,0 +1,269 @@
...
+
+    <ScrollView>
+        <StackPanel Orientation="Vertical" Padding="30,15">
+            <!--
+                We use separate grids for the header and each child because WinUI 3
+                doesn't support having a dynamic row count.
+
+                This unfortunately means we need to copy the resources and the
+                column definitions to each Grid.
+            -->
```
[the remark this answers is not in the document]
deansheather: > This would be more readable and modular if we componentized into multiple controls: header, sync session row, add new.

Each part could be moved into a `UserControl` but:
1. They'd only be used once
2. There are serious data binding issues I've encountered with some of the `UserControl`s we already have, which makes me think it won't be as easy as just copying/pasting and adding some properties

So although the file would be easier to work with potentially it would probably result in other issues unfortunately

> Also, isn't there some way to define the column definitions once and reference it?

Other than doing it in code for all three grids there's no way to do it once.

App/Models/MutagenSessionModel.cs
```diff
@@ -0,0 +1,310 @@
+using System;
+using Coder.Desktop.App.Converters;
+using Coder.Desktop.MutagenSdk.Proto.Synchronization;
+using Coder.Desktop.MutagenSdk.Proto.Url;
+
+namespace Coder.Desktop.App.Models;
+
+// This is a much slimmer enum than the original enum from Mutagen and only
+// contains the overarching states that we care about from a code perspective.
+// We still store the original state in the model for rendering purposes.
```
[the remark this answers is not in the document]
deansheather: > Do you think we are really going to use these "high level" statuses generically in the code beyond just choosing the color?

Probably not

> I'm tempted to say we should just stick with the mutagen statuses as they are and have a function in the view that converts them to a color.

I disagree. I'm going to rename this class to `SyncSessionModel` shortly because it shouldn't be coupled to Mutagen or use any of the protobuf types. The mutagen statuses have no meaning beyond their strings being shown to the user.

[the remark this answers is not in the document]
deansheather: If we really did want to use mutagen statuses we would also need a giant `DependencyObjectSelector` for color mapping which offers no compile time checks unfortunately, which makes me want to avoid using it for large enums

App/ViewModels/FileSyncListViewModel.cs
deansheather: Not an option unfortunately :(

App/ViewModels/FileSyncListViewModel.cs
deansheather: Users will not be able to open the window if the VPN is not active. This handles closing the window when the VPN session is closed.

App/Views/FileSyncListWindow.xaml
deansheather: The mockup says `Coder Desktop`, but I agree it should probably be different

App/Views/FileSyncListWindow.xaml
deansheather: In this case I'm only using it for minimum window size setting. We might start using it in other places, but probably as necessary rather than as a global thing. `WindowEx` inherits `Window` so it's pretty much a drop-in replacement when required

App/Views/Pages/TrayWindowMainPage.xaml
deansheather: The reason I added `DependencyObjectSelector` was for being able to convert from an enum to a `ThemeResource` color, which is necessary in `FileSyncListMainPage` because we're mapping to WinUI resource colors that change when the system theme changes.

In the case of `AgentViewModel.ConnectionStatus`, we aren't using theme resources and instead using hardcoded colors which look OK on both light and dark themes (and probably not high contrast themes).

Since we're using hardcoded colors we could keep the converter, but it seems weird to use one method in one place and another in a different place. I agree it's less readable, though.

Also, if we decide to use theme resources for connection status (which we probably should TBH) we will need to switch to this method anyways

App/Converters/DependencyObjectSelector.cs
deansheather: I can add better comments, but...

`DependencyObjectSelector` contains multiple `DependencyObjectSelectorItem`s, each with a key and a value. The `DependencyObjectSelector` has a dependency property on it called `SelectedKey`, which when updated from the view will cause `SelectedObject` to be updated to the value of the matching `DependencyObjectSelectorItem`

spikecurtis: Ah! If I'm understanding you correctly, this is a map, but for DependencyProperties so it does the right stuff to wire up to the presentation framework. And, the reason a Converter won't work is that the converter approach only triggers updates when the "key" side gets updated, whereas for themed colors, the "values" could also update on a theme change, and we want to propagate that change thru.

spikecurtis: This class could really use a unit test that updates both keys and values and asserts that the changes propagate.

App/Views/Pages/TrayWindowMainPage.xaml
spikecurtis: Is there some way we could instantiate this resource in plain C# ? I feel like not paying the [Angle Bracket Tax](https://blog.codinghorror.com/xml-the-angle-bracket-tax/) alone would help a lot.

App/ViewModels/FileSyncListViewModel.cs
spikecurtis: Preventing them from opening it is fine if it's reasonably clear why, as is closing it if it is in response to some _action_ the user has taken that can be reasonably seen as related.

It looks to me like we would close the window if the VPN service crashed or otherwise went down other than via user action.

App/Models/MutagenSessionModel.cs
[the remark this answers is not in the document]
spikecurtis: I like renaming to SyncSessionModel. 

Instead of a large Enum of mutagen statuses, can we rename this to something that is conceptually distinct from "status". It's confusing to have both Status and StatusString and have different content. I hesitate to reference "color" in a model since that's visual representation, but maybe "status category"?

App/Views/Pages/FileSyncListMainPage.xaml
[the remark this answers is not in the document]
spikecurtis: > Other than doing it in code for all three grids there's no way to do it once.

Coming from React it surprises me how unergonomic some of this XAML stuff is.

App/ViewModels/FileSyncListViewModel.cs
deansheather: That behavior sounds fine to me. The user won't be able to add new sync sessions since we won't have a list of agents to show in the list, so the UI will already have reduced functionality anyways

App/Models/MutagenSessionModel.cs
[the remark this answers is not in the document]
deansheather: To me it sounds more like something you'd configure with the CLI or dashboard first, rather than the desktop app being the main way to configure it

In a workspace <=> workspace sync, either alpha or beta would be a local protocol and one would be SSH, so they wouldn't both be SSH targets (and we would need to handle that in the UI a lot differently I think)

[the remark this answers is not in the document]
deansheather: But yeah I agree with Spike, I'll undo the alpha/beta => local/remote replacements

App/ViewModels/FileSyncListViewModel.cs
spikecurtis: It's not about what the user can and can't reasonably do---it's about whether they can understand why things are happening. Think about it from the user's perspective: if you're setting up a file sync session and the window just suddenly closes, what do you think? Probably that the app crashed; that it's glichy and buggy and not to be trusted.

deansheather: Would you prefer if the window had a "unavailable" state?

spikecurtis: Just having it would be an improvement, since it would look less like things crashed. If it explained why its unavailable that would be even better.

App/Views/Pages/TrayWindowMainPage.xaml
deansheather: Discussed on Slack, it doesn't seem to be possible or I cannot figure out how to do it. :/

App/Converters/DependencyObjectSelector.cs
deansheather: Discussed on Slack, due to issues it's difficult to test this currently: #62

App/ViewModels/FileSyncListViewModel.cs
deansheather: Added some text that shows when file syncs aren't available based on the RpcLifecycle, VpnLifecycle or CredentialState

App/Services/MutagenController.cs
```diff
@@ -132,11 +122,10 @@
             _sessionCount += 1;
         }
 
-        return session;
+        throw new NotImplementedException();
```
spikecurtis, line 125: This breaks the UTs --- and doesn't seem to be related to the UI view. Did it get added by accident to this PR?

App/ViewModels/FileSyncListViewModel.cs
```diff
@@ -0,0 +1,253 @@
...
+public partial class FileSyncListViewModel : ObservableObject
+{
+    private DispatcherQueue? _dispatcherQueue;
+
+    private readonly ISyncSessionController _syncSessionController;
+    private readonly IRpcController _rpcController;
+    private readonly ICredentialManager _credentialManager;
+
+    [ObservableProperty]
+    [NotifyPropertyChangedFor(nameof(ShowUnavailable))]
```
spikecurtis, line 26: ShowUnavailable doesn't depend on Loading.

App/ViewModels/FileSyncListViewModel.cs
```diff
@@ -0,0 +1,253 @@
...
+    [NotifyPropertyChangedFor(nameof(ShowLoading))]
+    [NotifyPropertyChangedFor(nameof(ShowError))]
```
spikecurtis, line 28: ShowError doesn't depend on Loading

App/ViewModels/FileSyncListViewModel.cs
```diff
@@ -0,0 +1,253 @@
...
+    [NotifyPropertyChangedFor(nameof(ShowSessions))]
+    public partial string? UnavailableMessage { get; set; } = null;
+
+    [ObservableProperty]
+    [NotifyPropertyChangedFor(nameof(ShowUnavailable))]
```
spikecurtis, line 40: ShowUnavailable doesn't depend on Error.

App/ViewModels/FileSyncListViewModel.cs
```diff
@@ -0,0 +1,253 @@
...
+            new SyncSessionModel(@"C:\Users\dean\git\coder", "pog", "~/coder", SyncSessionStatusCategory.Unknown,
+                "Unknown", "Some description", []),
+            new SyncSessionModel(@"C:\Users\dean\git\coder", "pog", "~/coder", SyncSessionStatusCategory.Working,
+                "Reconciling", "Some description", []),
+        ];
+    }
+
+    public void Initialize(DispatcherQueue dispatcherQueue)
+    {
+        _dispatcherQueue = dispatcherQueue;
```
spikecurtis, line 112: do we also need to check `HasThreadAccess`?

App/Models/SyncSessionModel.cs
```diff
@@ -0,0 +1,249 @@
...
+
+    public SyncSessionModel(State state)
+    {
+        Identifier = state.Session.Identifier;
+        Name = state.Session.Name;
+
+        // If the protocol isn't what we expect for alpha or beta, show
+        // "unknown".
+        if (state.Session.Alpha.Protocol == Protocol.Local && !string.IsNullOrWhiteSpace(state.Session.Alpha.Path))
+            LocalPath = state.Session.Alpha.Path;
```
spikecurtis, line 116: Looks like we're still assuming alpha is always local and beta is always remote.

App/Models/SyncSessionModel.cs
```diff
@@ -0,0 +1,249 @@
...
+
+// This is a much slimmer enum than the original enum from Mutagen and only
+// contains the overarching states that we care about from a code perspective.
+// We still store the original state in the model for rendering purposes.
+public enum SyncSessionStatusCategory
+{
+    Unknown,
+    Paused,
+    Error,
+    Conflicts,
```
spikecurtis, line 16: Worth a comment that categories higher than conflicts will get replaced by conflicts if there are any.

App/ViewModels/FileSyncListViewModel.cs
deansheather: This is only being called in one place in a main thread, I could add a check for that and throw an exception if you want.

spikecurtis: Check and throw is good. Want to avoid little landmines on the public interface of this class.

App/Services/MutagenController.cs
deansheather: My bad, with the type changes I made I had to change the output of this function and wasn't thinking about UTs lol.

spikecurtis (approved):

PR #64, opened 2025-03-26
feat: wire up file sync window

### LINKED ISSUE #26: Add basic file sync support using GUI with manually entered paths
Add basic support to sync directories with a GUI to enter sync paths and view existing file syncs. Users must enter paths manually; no GUI at this stage.

Image

## Not in scope

 * Metadata about existing syncs like status, size, etc
 * GUI for picking folders

## Dependencies

 * coder/internal#378
 * coder/coder#16490
 * #25 
 * coder/internal#380

### LINKED ISSUE #28: Add local folder picker to file sync GUI
Add a local folder picker to the file sync GUI. This makes it so that you can visually browse the file system right in the application, rather than having to open the native Explorer.

Image

## Dependencies

 * #26

### LINKED ISSUE #29: Add real time status metadata to file sync GUI
Adds a real time metadata columns to the file sync GUI, based on the [state](https://github.com/mutagen-io/mutagen/blob/master/pkg/synchronization/state.proto) returned by the mutagen daemon.

Should include:

* Status
* Number of files
* Total size

## Dependencies

 * #26

deansheather:
- Adds `PauseSyncSession` and `ResumeSyncSession`
- Adds `SyncSessionViewModel` that wraps `SyncSessionModel` and adds view methods (as you cannot access the parent context if you're in a `ItemsRepeater` apparently)
- Wires up Initialize, List, Pause, Resume, Terminate and Create in the file sync UI

## TODO:
- Prevent the app from loading until mutagen finishes initializing (either successfully or not) (in a different PR)
- Add reinitialization logic to mutagen controller (in a different PR)

Closes #26 
Closes #28 
Closes #29

### DIFF
--- a/App/Models/SyncSessionModel.cs
+++ b/App/Models/SyncSessionModel.cs
@@ -51,6 +51,7 @@ public sealed class SyncSessionModelEndpointSize
 public class SyncSessionModel
 {
     public readonly string Identifier;
+    public readonly DateTime CreatedAt;
 
     public readonly string AlphaName;
     public readonly string AlphaPath;
@@ -99,6 +100,7 @@ public class SyncSessionModel
     public SyncSessionModel(State state)
     {
         Identifier = state.Session.Identifier;
+        CreatedAt = state.Session.CreationTime.ToDateTime();
 
         (AlphaName, AlphaPath) = NameAndPathFromUrl(state.Session.Alpha);
         (BetaName, BetaPath) = NameAndPathFromUrl(state.Session.Beta);
--- a/App/Models/SyncSessionModel.cs
+++ b/App/Models/SyncSessionModel.cs
@@ -1,4 +1,3 @@
-using System;
 using Coder.Desktop.App.Converters;
 using Coder.Desktop.MutagenSdk.Proto.Synchronization;
 using Coder.Desktop.MutagenSdk.Proto.Url;
@@ -64,6 +63,10 @@ public class SyncSessionModel
 
     public readonly string[] Errors = [];
 
+    // If Paused is true, the session can be resumed. If false, the session can
+    // be paused.
+    public bool Paused => StatusCategory is SyncSessionStatusCategory.Paused or SyncSessionStatusCategory.Halted;
+
     public string StatusDetails
     {
         get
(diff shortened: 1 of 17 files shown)


App/Models/SyncSessionModel.cs
```diff
@@ -45,6 +49,159 @@ public string Description(string linePrefix = "")
...
+    Unknown,
+    Directory,
+    File,
+    SymbolicLink,
+    Untracked,
+    Problematic,
+    PhantomDirectory,
+}
+
+public sealed class SyncSessionModelEntry
```
spikecurtis, line 63: This structure follows the mutagen message types more or less exactly. I understand the desire to keep our implementation relatively independent, so that we could conceivably replace Mutagen with something else down the line, but a message that looks like mutagen and quacks like mutagen might as well be mutagen.

It looks like for now, all we do with conflicts is convert them into a string description. Let's make that the abstraction that we make public, instead of locking in all this detail. We can generate the description on construction directly from the proto types, and it's easy to compare it for changes.

Later when we want to do something better on the UX, we can design a backing type that is responsive to that UX. That way we are only baking in the UX decisions we are making, rather than UX _and_ mutagen type heirarchy in everything but name.

Tests.App/Services/MutagenControllerTest.cs
```diff
@@ -37,6 +37,15 @@ public void CreateTempDir()
         TestContext.Out.WriteLine($"temp directory: {_tempDirectory}");
     }
 
+    [TearDown]
+    public void DeleteTempDir()
```
spikecurtis, line 41: Do we really want to do this? Makes it harder to debug. I assumed the OS would clear the directories out for us eventually, like on restart

App/Models/SyncSessionModel.cs
```diff
@@ -62,14 +219,22 @@ public class SyncSessionModel
     public readonly SyncSessionModelEndpointSize AlphaSize;
     public readonly SyncSessionModelEndpointSize BetaSize;
 
-    public readonly string[] Errors = [];
+    public readonly IReadOnlyList<SyncSessionModelConflict> Conflicts;
+    public ulong OmittedConflicts;
```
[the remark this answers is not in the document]
deansheather: Yep, good catch!

App/Services/MutagenController.cs
```diff
@@ -3,36 +3,67 @@
...
+using DaemonTerminateRequest = Coder.Desktop.MutagenSdk.Proto.Service.Daemon.TerminateRequest;
+using SynchronizationTerminateRequest = Coder.Desktop.MutagenSdk.Proto.Service.Synchronization.TerminateRequest;
 
 namespace Coder.Desktop.App.Services;
 
 public class CreateSyncSessionRequest
 {
-    // TODO: this
+    public Uri Alpha { get; init; }
+    public Uri Beta { get; init; }
```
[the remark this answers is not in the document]
deansheather: TBH IDK why I went this way originally, I agree that a struct would be better

Tests.App/Services/MutagenControllerTest.cs
deansheather: They get persisted unfortunately. I'll make it so it only deletes if the test passes

App/Services/MutagenController.cs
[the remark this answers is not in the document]
deansheather: Fixed

App/Models/SyncSessionModel.cs
deansheather: Fixed, just storing strings now

App/Services/MutagenController.cs
```diff
@@ -3,36 +3,83 @@
...
 namespace Coder.Desktop.App.Services;
 
-public class CreateSyncSessionRequest
+public enum CreateSyncSessionRequestEndpointProtocol
+{
+    Local,
+    Ssh,
+}
+
+public class CreateSyncSessionRequestEndpoint
```
spikecurtis, line 32: Just wondering out loud: these names are getting really long... is it C# style to namespace them rather than have a long prefix?

spikecurtis (approved):
LGTM!

App/Services/MutagenController.cs
deansheather: I'm not sure what the convention is. Another option would be to move these inside the "parent type" e.g. `new CreateSyncSessionRequest.Endpoint()`.

spikecurtis: That makes sense to me since they are only used to fully define the `CreateSyncSessionRequest` type

deansheather: Renamed these to:
- `CreateSyncSessionRequest.Endpoint`
- `CreateSyncSessionRequest.Endpoint.ProtocolKind` (can't be `Protocol` or it conflicts with the property with the same name)

So the names are longer when using them, but they're segregated and not cluttering up the main namespace

spikecurtis (approved):

Issue #52, opened 2025-03-11
RDP-specific Coder Desktop deep link


ThomasK33:
Add RDP-specific logic into Coder Desktop so that all coder apps marked with `external=true` and a URL matching to `coder:/v0/open/agent/<agent ID>/rdp?username=<username>&password=<passwpord>` will be rendered as buttons and have RDP-specific logic in place. This logic can include fetching additional information or credentials in Coder Desktop and creating an `.rdp` file or comparable integrations.

spikecurtis:
Thinking about this a bit more, I think the URI format should be:

`coder://coder.example.com/v0/open/ws/<workspace-name>/agent/<agent-name>/rdp?username=<username>&password=<password>`

We _should_ include the authority, so that if you have multiple Coder deployments and you click a link from one of them, and Coder Desktop is signed into a different one, the you can get a sensible error.

We should prefer name references over UUIDs, so that it's possible to bookmark. If we used Agent UUIDs, the link would break as soon as we rebuild.

matifali:
> We should include the authority, so that if you have multiple Coder deployments and you click a link from one of them, and Coder Desktop is signed into a different one, the you can get a sensible error.


This is a good addition. It would help provide better self-troubleshooting.

PR #72, opened 2025-04-22
feat: add support for URI activations for coder scheme

spikecurtis:
relates to #52

Adds support for Coder Desktop to handle the `coder:/` URI scheme by registering for this scheme and forwarding activations to the single instance of Coder Desktop.

Also removes the `Package.appxmanifest`. It is unused since Coder Desktop is not a packaged app.


spikecurtis:
* **#72** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/72?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/72?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

App/Program.cs
```diff
@@ -26,7 +27,23 @@ private static void Main(string[] args)
...
+            }
+
+            // Register for URI handling (known as "protocol activation")
+#if DEBUG
+            const string scheme = "coder-debug";
+#else
+            const string scheme = "coder";
+#endif
+            var thisBin = Assembly.GetExecutingAssembly().Location;
+            ActivationRegistrationManager.RegisterForProtocolActivation(scheme, thisBin + ",1", "Coder Desktop", "");
```
deansheather, line 45: Is it typical for apps to register themselves? Or is it more common for it to be registered in the installer (not sure if this is even an option)?

deansheather (approved):
Looks good!

App/Program.cs
spikecurtis: The installer could do it, but I'm not sure it provides any advantage.

One advantage of the App registering itself is that you don't need to install to debug, which is nice.

We might want to cosider _unregistering_ in the uninstaller, which I can look into in another PR.

PR #74, opened 2025-04-23
feat: add create progress to file sync window

deansheather:
https://github.com/user-attachments/assets/96d57799-0db8-4146-a93c-c157645cb4e9


deansheather:
Added a progress spinner too

image

spikecurtis (approved):
Looks good. 

Doesn't have to be in this PR, but can we make the Local Path, Workspace, and Remote Path fields be visually and actually uneditable during sync startup? It looks like you change their value while startup is happening, which wouldn't be good.

deansheather:
Tracking the uneditable fields in #82

PR #78, opened 2025-04-24
feat: add logging to App

spikecurtis:
Adds logging to the Coder Desktop application.

### DIFF
--- a/App/App.xaml.cs
+++ b/App/App.xaml.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Coder.Desktop.App.Models;
@@ -19,6 +18,7 @@ using Microsoft.Windows.AppLifecycle;
 using Windows.ApplicationModel.Activation;
 using Microsoft.Extensions.Logging;
 using Serilog;
+using System.Collections.Generic;
 
 namespace Coder.Desktop.App;
 
@@ -28,9 +28,6 @@ public partial class App : Application
 
     private bool _handleWindowClosed = true;
     private const string MutagenControllerConfigSection = "MutagenController";
-
-    private const string logTemplate =
-        "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {SourceContext} - {Message:lj}{NewLine}{Exception}";
 
 #if !DEBUG
     private const string ConfigSubKey = @"SOFTWARE\Coder Desktop\App";
@@ -45,9 +42,15 @@ public partial class App : Application
     public App()
     {
         var builder = Host.CreateApplicationBuilder();
+        var configBuilder = builder.Configuration as IConfigurationBuilder;
 
-        (builder.Configuration as IConfigurationBuilder).Add(
+        // Add config in increasing order of precedence: first builtin defaults, then HKLM, finally HKCU
+        // so that the user's settings in the registry take precedence.
+        AddDefaultConfig(configBuilder);
+        configBuilder.Add(
             new RegistryConfigurationSource(Registry.LocalMachine, ConfigSubKey));
+        configBuilder.Add(
+            new RegistryConfigurationSource(Registry.CurrentUser, ConfigSubKey));
 
         var services = builder.Services;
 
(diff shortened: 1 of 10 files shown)


spikecurtis:
* **#78** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/78?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/78?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Installer/Program.cs
```diff
@@ -266,12 +270,32 @@ private static int BuildMsiPackage(MsiOptions opts)
...
+                                "MutagenController": {
+                                    "MutagenExecutablePath": "[INSTALLFOLDER]mutagen.exe"
+                                },
+                                "Serilog": {
+                                    "Using": ["Serilog.Sinks.File"],
+                                    "MinimumLevel": "Information",
+                                    "Enrich": [ "FromLogContext" ],
+                                }
+                            }
+                            """;
```
spikecurtis, line 289: if we're cool with this style of configuration, I can also apply it to the VPN service.

App/App.xaml.cs
```diff
@@ -24,22 +27,51 @@ public partial class App : Application
...
 
+        // Logging
+        builder.Services.AddSerilog((_, loggerConfig) =>
+        {
+            loggerConfig.ReadFrom.Configuration(builder.Configuration);
+            var sinkConfig = builder.Configuration.GetSection("Serilog").GetSection("WriteTo");
+            if (!sinkConfig.GetChildren().Any())
+            {
+                // no log sink defined in the registry, so we'll add one here.
+                // We can't generally define these in the registry because we don't
```
deansheather, line 62: We should add a new registry key under HKEY_CURRENT_USER instead of adding it to the machine-wide config

Installer/Program.cs
```diff
@@ -266,12 +270,32 @@ private static int BuildMsiPackage(MsiOptions opts)
...
-                @"[INSTALLFOLDER]mutagen.exe")
+            new RegValue(RegistryHive, RegistryKey, "Manager:TunnelBinaryAllowVersionMismatch", "false")
         );
 
+        // rather than adding individual reg values, it's much clearer to write the whole config as JSON then
+        // programmatically spit out the registry values.
+        var appConfigJson = """
+                            {
+                                "MutagenController": {
+                                    "MutagenExecutablePath": "[INSTALLFOLDER]vpn\mutagen.exe"
```
deansheather, line 281:
```suggestion
                                    "MutagenExecutablePath": "[INSTALLFOLDER]vpn\\mutagen.exe"
```

App/App.xaml.cs
```diff
@@ -24,22 +27,51 @@ public partial class App : Application
...
+        "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {SourceContext} - {Message:lj}{NewLine}{Exception}";
 
 #if !DEBUG
-    private const string MutagenControllerConfigSection = "AppMutagenController";
+    private const string ConfigSubKey = @"SOFTWARE\Coder Desktop\App";
+    private const string logFilename = "app.log";
 #else
-    private const string MutagenControllerConfigSection = "DebugAppMutagenController";
+    private const string ConfigSubKey = @"SOFTWARE\Coder Desktop\DebugApp";
+    private const string logFilename = "debug-app.log";
```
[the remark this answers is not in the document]
spikecurtis: We can do `Coder Desktop Debug\App` instead of `Coder Desktop\DebugApp` --- but I do want `App` to be its own key to keep it distinct from the VPN service. Things like `Serilog` have defined config they understand, and I think we will want to have separate control over the app and the VPN service.

App/App.xaml.cs
spikecurtis: I'm not sure that helps---the registry keys get written by the installer, and I don't think we can count on the fact that the user that runs the installer will be the user that runs Coder Desktop.

App/App.xaml.cs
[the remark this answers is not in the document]
deansheather: Yeah I didn't put two and two together about the serilog stuff, so your way is better. I don't particularly mind what key we use then, but the service keys should probably be fixed in this PR too to match the new format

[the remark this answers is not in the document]
deansheather: Any log configuration should go in `HKEY_CURRENT_USER` though IMO

App/App.xaml.cs
deansheather: I'm suggesting that the installer shouldn't write anything to any user-specific keys, and the app should just default to these settings instead when there's nothing configured

App/App.xaml.cs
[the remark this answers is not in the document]
spikecurtis: The Mutagen binary path still kinda needs to live in HKLM, since it's set by the installer. I've made it so that it starts with default config, hardcoded into the app, then checks HKLM, then HKCU.

App/App.xaml.cs
```diff
@@ -126,9 +151,14 @@ protected override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventArgs ar
         _ = syncSessionController.RefreshState(syncSessionCts.Token).ContinueWith(t =>
         {
             // TODO: log
```
deansheather, line 153:
```suggestion
```

App/App.xaml.cs
```diff
@@ -110,13 +133,15 @@ protected override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventArgs ar
         _ = credentialManager.LoadCredentials(credentialManagerCts.Token).ContinueWith(t =>
         {
             // TODO: log
```
deansheather, line 135:
```suggestion
```

deansheather (approved):

spikecurtis:
### Merge activity

* **Apr 30, 7:25 AM EDT**: @spikecurtis merged this pull request with [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-windows/78).

PR #81, opened 2025-04-29
fix: fix Downloader to dispose tempFile and use synchronous IO 

spikecurtis:
Fixes https://github.com/coder/internal/issues/598

There is a possible race where if the cancellation token is expired, `Download()` never gets called and the tempFile is never disposed of (at least until GC). We also switch to synchronous IO so that a pending overlapped write won't block the deletion.

These issues can cause races in our tests when we try to clean up the directory.

### DIFF
--- a/Tests.Vpn.Service/DownloaderTest.cs
+++ b/Tests.Vpn.Service/DownloaderTest.cs
@@ -442,23 +442,28 @@ public class DownloaderTest
     [CancelAfter(30_000)]
     public async Task CancelledInner(CancellationToken ct)
     {
+        var httpCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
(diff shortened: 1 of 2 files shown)


spikecurtis:
* **#81** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/81?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/81?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

deansheather (approved):

spikecurtis:
Hmm. Still getting the error, so clearly the issue I noticed must not be the (only) cause.

PR #84, opened 2025-05-01
fix: ensure mutagen daemon log is closed

deansheather:


App/Services/MutagenController.cs
```diff
@@ -556,25 +556,61 @@ private void StartDaemonProcess()
...
+            try
+            {
+                daemonProcess.Kill();
+            }
+            catch
+            {
+                // ignored, the process likely doesn't exist
+            }
+
+            daemonProcess.Dispose();
```
spikecurtis, line 610: should re-throw, right?

deansheather: oops

spikecurtis (approved):

PR #85, opened 2025-05-01
feat: add support for notifications

spikecurtis:
Adds support for OS notifications, which I'll use to show errors handling URIs in a subsequent PR.

[Screen Recording 2025-05-01 145532.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/media/video/tCz4CxRU9jhAJ7zH8RTi/f838fb8a-6815-48a7-bd52-63d6a06ce742.mp4)

### DIFF
--- a/App/Services/UserNotifier.cs
+++ b/App/Services/UserNotifier.cs
@@ -1,0 +1,29 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.Windows.AppNotifications;
+using Microsoft.Windows.AppNotifications.Builder;
+
+namespace Coder.Desktop.App.Services;
+
+public interface IUserNotifier : IAsyncDisposable
+{
+    public Task ShowErrorNotification(string title, string message);
+}
+
+public class UserNotifier : IUserNotifier
+{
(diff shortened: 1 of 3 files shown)


App/Services/UserNotifier.cs
```diff
@@ -0,0 +1,29 @@
...
+public class UserNotifier : IUserNotifier
+{
+    private readonly AppNotificationManager _notificationManager = AppNotificationManager.Default;
+
+    public ValueTask DisposeAsync()
+    {
+        return ValueTask.CompletedTask;
+    }
+
+    public Task ShowErrorNotification(string title, string message)
```
deansheather, line 22: This could just be `ShowNotification`, the function itself doesn't seem to be only relevant for errors. I imagine we could use it for other notifications like workspace lifetime notifications in the future

deansheather (approved):

App/Services/UserNotifier.cs
spikecurtis: I kinda want to keep the information that this is an "error" in case we want to refactor this class to use, say, a dialog pop-up instead, or decide to style the notifications further to make errors stand out from other notifications.

PR #86, opened 2025-05-06
feat: add workspace app icons to tray window

### LINKED ISSUE #50: Button Auto-Population for External Apps
Enable Coder Desktop to detect and display buttons for external applications automatically. Users desire a seamless, one-click experience for launching native applications such as RDP clients, Zed, Visual Studio Code, and more. This feature will allow Coder Desktop to show relevant buttons based on the configuration of the workspace template.

Initially, we want to display all `coder_app`'s marked with `external=true` and a set `url` field that does not have an `http` or `https` scheme. (e.g., `zed://`, `rdp://`, etc.) as buttons.

This information can be obtained by retrieving the workspace's agent, enumerating all applications, and filtering Coder Desktop-side. When the button is clicked, it should delegate the task of opening the link to the operating system.

To keep the user interface simple and intuitive, we will display a maximum of five buttons, creating a consistent row height. In future iterations, we can explore other options, such as dropdown menus or collapsible sections.

deansheather:
- Adds `AgentAppViewModel` to handle each button
- Adds collapsible control components to handle the collapsing section
- Adds `Uuid` type to work around issues with the built-in `Guid` type
- Adds `ModelMerge` utility for merging lists with minimal updates to work around constant flashing in the UI

### TODO:
- [x] Auto expand the first agent (to match macOS)
- [x] Only allow one agent to be expanded at a time (to match macOS)
- [x] Add tests for `ModelMerge`
- [x] Get rid of the SVG size constants in `AgentAppViewModel` since I don't think they do anything
- [x] Add `$SESSION_TOKEN` replacement
- [x] Display icons for vscode and vscode insiders

### Known issues:
- `/icon/cursor.svg` seems to be an SVG with an embedded PNG, which I guess can't be rendered properly by WinUI

https://github.com/user-attachments/assets/b1f2b1d9-7468-4288-afef-b73d9db5252c

Closes #50


spikecurtis (commented):
I didn't finish looking at all the files, but it looks like you're still doing some stuff, so I'll get another look later anyway. It's the end of my day so sending the comments I have so you'll have them to look at in the morning.

App/ViewModels/AgentViewModel.cs
```diff
@@ -14,17 +65,228 @@ public enum AgentConnectionStatus
...
+        AppFetchErrored = !task.IsCompletedSuccessfully;
+        if (!task.IsCompletedSuccessfully)
+        {
+            _logger.LogWarning(task.Exception, "Could not fetch workspace agent");
+            return;
+        }
+
+        var workspaceAgent = task.Result;
+        var apps = new List<AgentAppViewModel>();
+        foreach (var app in workspaceAgent.Apps)
```
ethanndickson, line 262: I didn't see it in the TODOs, but I had to render `display_apps` for VS Code and VS Code Insiders apps.

App/Utils/ModelMerge.cs
```diff
@@ -0,0 +1,85 @@
...
+///     ItemsRepeater items when nothing has changed.
+/// </summary>
+public static class ModelMerge
+{
+    /// <summary>
+    ///     Merges two observable lists with as little operations as possible
+    ///     to avoid excessive/unncessary UI updates.
+    ///     It's assumed that the target list is already sorted.
+    /// </summary>
+    public static void MergeLists<T>(IList<T> target, IList<T> update, Comparison<T> sorter)
```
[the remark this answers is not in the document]
spikecurtis: > You should actually only get [A, B, E] after that merge, but using the existing A reference after calling ApplyMerge on it.

I'm saying that the term "merge", to me, makes it _seem_ like you should get all elements of both lists. Like, when you merge a PR, you don't just get the code from one parent. The name is confusing to me; the actual behavior of this code is fine.

Aaanyway. I think we're on the same page, I just didn't express myself as clearly as I could have.

App/ViewModels/AgentAppViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+            {
+                UseFallbackIcon = true;
+                return new BitmapImage();
+            }
+
+            // Determine what image source to use based on extension, use a
+            // BitmapImage as last resort.
+            var ext = IconUrl.AbsolutePath.Split('/').LastOrDefault()?.Split('.').LastOrDefault();
+            // TODO: this is definitely a hack, URLs shouldn't need to end in .svg
+            if (ext is "svg")
```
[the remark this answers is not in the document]
spikecurtis: Agree that we can add an issue and track it for GA.

App/Utils/ModelMerge.cs
```diff
@@ -0,0 +1,85 @@
...
+public interface IModelMergeable<in T>
+{
+    /// <summary>
+    ///     Applies a merge of obj to `this`.
+    /// </summary>
+    /// <returns>
+    ///     True if the two objects represent the same item (e.g. the ID/name
+    ///     matches) and the merge was performed.
+    /// </returns>
+    public bool ApplyMerge(T obj);
```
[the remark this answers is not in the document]
deansheather: `TryApplyChanges` is what I went with it

App/Utils/ModelMerge.cs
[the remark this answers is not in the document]
deansheather: `ModelUpdate.ApplyLists` is what I went with.

App/Utils/ModelMerge.cs
```diff
@@ -0,0 +1,85 @@
...
+            i--;
+
+            OuterLoopEnd: ;
+        }
+
+        // Add any items that were missing into their correct sorted place.
+        // It's assumed the list is already sorted.
+        foreach (var item in newItems)
+        {
+            // Perform a binary search. List<T> has BinarySearch(), but
```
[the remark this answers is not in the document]
deansheather: Changed to linear search

App/ViewModels/AgentAppViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+    public string Details =>
+        (string.IsNullOrWhiteSpace(Name) ? "(no name)" : Name) + ":\n\n" + AppUri;
+
+    public ImageSource ImageSource
+    {
+        get
+        {
+            if (IconUrl is null || (IconUrl.Scheme != "http" && IconUrl.Scheme != "https"))
+            {
+                UseFallbackIcon = true;
```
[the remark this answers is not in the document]
deansheather: It'd probably be more idiomatic to use a new ObservableProperty for the ImageSource and use a `PropertyChanged` event listener to update it when the icon URL changes.

App/ViewModels/AgentAppViewModel.cs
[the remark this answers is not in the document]
deansheather: #91, #92

App/ViewModels/AgentViewModel.cs
deansheather: Added them and added issues for using coder-remote: #93 and https://github.com/coder/coder-desktop-macos/issues/156

App/ViewModels/TrayWindowViewModel.cs
```diff
@@ -156,75 +163,101 @@ private void UpdateFromRpcModel(RpcModel rpcModel)
...
+                    _hasExpandedAgent = true;
+                    // Collapse every other agent.
+                    foreach (var otherAgent in Agents.Where(a => a.Id != agent.Id && a.IsExpanded))
+                        otherAgent.IsExpanded = false;
+                }
+            };
 
         // Sort by status green, red, gray, then by hostname.
-        agents.Sort((a, b) =>
+        ModelUpdate.ApplyLists(Agents, agents, (a, b) =>
```
spikecurtis, line 211: I don't think this application will propagate the `PropertyChanged` event, which means that existing agents won't collapse new agents when they are expanded.

Having agent view models collapse other agent view models feels fragile. Could we have some state on this view model that controls which agent is expanded, rather than a bunch of IsExpanded bools on each agent view model that we need to toggle? Or, could we lift the routine that expands an agent up to this model, so that it can expand exactly one of them?

CoderSdk/Coder/WorkspaceAgents.cs
```diff
@@ -0,0 +1,38 @@
...
+    public string Url { get; set; } = string.Empty;
+    public bool External { get; set; } = false;
+    public string DisplayName { get; set; } = string.Empty;
+    public string Command { get; set; } = string.Empty;
+    public string Icon { get; set; } = string.Empty;
+}
+
+public partial class CoderApiClient
+{
+    public Task<WorkspaceAgent> GetWorkspaceAgent(string id, CancellationToken ct = default)
```
spikecurtis, line 34: This would be safer if we accepted a UUID, rather than a string, which could totally change the endpoint we hit.

deansheather: I considered this but didn't want to add a dependency between `CoderSdk` and `Vpn.Proto`, and adding a new project for just a UUID type seems overkill. The Go codersdk package accepts string IDs for some methods as well, so I think it's fine.

App/ViewModels/TrayWindowViewModel.cs
ethanndickson: +1, on the mac app we just modify `isExpanded: UUID?` to do this.

App/ViewModels/AgentAppViewModel.cs
```diff
@@ -0,0 +1,177 @@
...
+
+    [RelayCommand]
+    private void OpenApp(object parameter)
+    {
+        try
+        {
+            var uriString = AppUri;
+            var cred = _credentialManager.GetCachedCredentials();
+            if (cred.State is CredentialState.Valid && cred.ApiToken is not null)
+                uriString = uriString.Replace(SessionTokenUriVar, cred.ApiToken);
```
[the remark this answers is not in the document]
deansheather: @ethanndickson you'll need to do this in macOS too I think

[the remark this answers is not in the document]
ethanndickson: the mac app just doesn't display `http` or `https` apps at all, actually. (On the basis those should always have `external` set to true)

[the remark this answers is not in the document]
deansheather: You can create external http/https apps in coder, there's no restriction against that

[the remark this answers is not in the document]
ethanndickson: Oh sorry, nah it was Atif's suggestion to filter out things like internal wikis and documentation pages some of our clients like to link via apps.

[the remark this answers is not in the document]
deansheather: I've added a filter for `http` and `https` apps to the code in `AgentViewModel` that populates them, and I've also avoided replacing `$SESSION_TOKEN` in any `http` or `https` URLs in the `AgentAppViewModel` as well in case the population behavior is changed.

App/ViewModels/TrayWindowViewModel.cs
deansheather: > I don't think this application will propagate the PropertyChanged event, which means that existing agents won't collapse new agents when they are expanded.

It won't propagate it, but every single agent that was created will get the event listener added exactly once when created.

spikecurtis: Oh, whoops, I missed that you are iterating over `Agents` not `agents`. You're right that each new agent gets the listener, but now there could be concurrent access to `Agents` while you are modifying them.

deansheather: There shouldn't be concurrent access either since we're doing all of this in the UI thread (`PropertyChanged` gets called immediately when setting an observable property, which can only happen in the UI thread), and all changes to `Agents` also happen in the UI thread.

I agree it's not pretty to do it this way, but I think the idea of passing down the ID of the currently expanded item to each child is also not pretty IMO. It works better in Swift since you can safely have a pointer to an ID shared between all `AgentViewModel`s, but you can't do that in C# without using `unsafe` or wrapping the ID in a class and using that as a reference. I thought about implementing it with a binding but since we're not creating these view models in the XAML code we have to write the bindings ourself which is not pretty.

Since we're using an `ItemsRepeater` to render these agents, the data context changes and we can't access the `TrayWindowViewModel` to e.g. set the command handler for the button to be a method on the parent view model _unless_ we pass `TrayWindowViewModel` to the child `AgentViewModel`, which is also not pretty.

Every method of doing this seems to suck in a different way unfortunately. The reason I went this way originally was so none of the `AgentViewModel`s had to be aware of any of the other `AgentViewModel`s.

deansheather: Spike and I discussed in a call and we've opted to:
1. Create a new interface `IAgentExpanderHost` which is implemented by `TrayWindowViewModel`
2. Pass the `TrayWindowViewModel` as an `IAgentExpanderHost` into all of the `AgentViewModel`s
3. Have each of the `AgentViewModel`s call the method on the interface when `this.IsExpanded` changes
4. The implementation on `TrayWindowViewModel` will collapse every other agent by calling `agent.SetExpanded(false)`

CoderSdk/Coder/WorkspaceAgents.cs
spikecurtis: I had (wrongly) assumed/remembered that we parsed the `Agent` protobufs into a model with `Guid` id types. But, we just leave it as a byte string. You're right, not worth introducing the dep.

spikecurtis: Actually, is there some reason we need to roll our own `Uuid` class? I believe https://learn.microsoft.com/en-us/dotnet/api/system.guid?view=net-9.0 is the same thing, but MSFT uses a different name than other systems.

spikecurtis: Could we move our `Uuid` implementation to CoderSdk? There will likely be plenty of times we want to interact with UUIDs outside of the protobuf protocols.

Vpn.Proto/Uuid.cs
```diff
@@ -0,0 +1,175 @@
...
+namespace Coder.Desktop.Vpn.Proto;
+
+/// <summary>
+///     A simplistic UUIDv4 class that wraps a 16-byte array. We don't use the
+///     native Guid class because it has some weird reordering behavior due to
+///     legacy Windows stuff. This class is not guaranteed to provide full RFC
+///     4122 compliance, but it should provide enough coverage for Coder
+///     Desktop.
+/// </summary>
+public class Uuid
```
spikecurtis, line 14: Can we move this to CoderSdk?

spikecurtis (approved):

spikecurtis:
I suggested we move `Uuid` to CoderSdk, but I don't need to review again.

PR #87, opened 2025-05-06
feat: add support for RDP URIs

spikecurtis:


spikecurtis:
* **#87** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/87?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/87?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

App/App.xaml.cs
```diff
@@ -190,7 +194,18 @@ public void OnActivated(object? sender, AppActivationArguments args)
...
+                try
+                {
+                    // don't need to wait for it to complete.
+                    _ = _uriHandler.HandleUri(protoArgs.Uri);
+                }
+                catch (System.Exception e)
+                {
+                    _logger.LogError(e, "unhandled exception while processing URI coder://{authority}{path}",
+                        protoArgs.Uri.Authority, protoArgs.Uri.AbsolutePath);
+                }
```
deansheather, line 207: This would probably be better as a `Task.ContinueWith()` call to log the exception

App/Services/RdpConnector.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    public readonly string Username = username;
+    public readonly string Password = password;
+}
+
+public interface IRdpConnector : IAsyncDisposable
+{
+    public const int DefaultPort = 3389;
+
+    public Task WriteCredentials(string fqdn, RdpCredentials credentials, CancellationToken ct = default);
```
deansheather, line 19: If this doesn't need to be async then it should probably just be synchronous for now. If we do need to change it to be async then we can do it pretty easily in the future.

App/Services/RdpConnector.cs
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Coder.Desktop.App.Services;
+
+public struct RdpCredentials(string username, string password)
+{
+    public readonly string Username = username;
+    public readonly string Password = password;
+}
+
+public interface IRdpConnector : IAsyncDisposable
```
deansheather, line 15: This doesn't need to be disposable for now. Easy to add later if needed

App/Services/UriHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+using System.Threading.Tasks;
+using System.Web;
+using Coder.Desktop.App.Models;
+using Coder.Desktop.Vpn.Proto;
+using Microsoft.Extensions.Logging;
+
+
+namespace Coder.Desktop.App.Services;
+
+public interface IUriHandler : IAsyncDisposable
```
deansheather, line 14: Same here, this also doesn't need to be disposable for now

App/Services/UriHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+
+public class UriHandler(
+    ILogger<UriHandler> logger,
+    IRpcController rpcController,
+    IUserNotifier userNotifier,
+    IRdpConnector rdpConnector) : IUriHandler
+{
+    private const string OpenWorkspacePrefix = "/v0/open/ws/";
+
+    internal class UriException(string title, string detail) : Exception
```
deansheather, line 27: AFAIK this creates an exception with no message. It's fine for your use case of showing a notification only, but it probably should still be a valid exception.

You'll need to create a proper constructor function and call the base constructor.

e.g.

```cs
internal class UriException : Exception
{
  // ...
  internal UriException(string title, string detail) : base($"{title}: {detail}")
  {
    Title = title;
    Detail = detail;
  }
}
```

App/Services/UriHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+        try
+        {
+            workspace = state.Workspaces.Single(w => w.Name == workspaceName);
+        }
+        catch (InvalidOperationException) // Single() throws this when nothing matches.
+        {
+            logger.LogDebug("got URI to open workspace {workspace}, but the workspace doesn't exist", workspaceName);
+            throw new UriException(errTitle,
+                $"Failed to open application on workspace {workspaceName} because it doesn't exist");
+        }
```
deansheather, line 91: A nicer pattern I've used that doesn't require a try/catch and predeclaring the variable outside the try block is:
```cs
var workspace = state.Workspaces.FirstOrDefault(w => w.Name == workspaceName);
if (workspace is null)
{
  // ...
  throw new Exception("");
}
```

App/Services/UriHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+        try
+        {
+            agent = state.Agents.Single(a => a.WorkspaceId == workspace.Id && a.Name == agentName);
+        }
+        catch (InvalidOperationException) // Single() throws this when nothing matches.
+        {
+            logger.LogDebug("got URI to open workspace/agent {workspaceName}/{agentName}, but the agent doesn't exist",
+                workspaceName, agentName);
+            throw new UriException(errTitle,
+                $"Failed to open application on workspace {workspaceName}, agent {agentName} because it doesn't exist.");
```
deansheather, line 103: The error message doesn't really specify what doesn't exist. Maybe try:

```
Failed to open application on workspace/agent {workspaceName}/{agentName} because the agent doesn't exist.
```

or something else

App/Services/UriHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+    {
+        if (uri.AbsolutePath.StartsWith(OpenWorkspacePrefix))
+        {
+            await HandleOpenWorkspaceApp(uri, ct);
+            return;
+        }
+
+        logger.LogWarning("unhandled URI path {path}", uri.AbsolutePath);
+        throw new UriException("URI handling error",
+            $"URI with path {uri.AbsolutePath} is unsupported or malformed");
```
deansheather, line 55: I think in a lot of these error messages it might be nice to surround the templated variables with single quotes. It can help users determine when a string that was supposed to be specified was empty, otherwise it's as if it's not in the string at all.

App/Services/UriHandler.cs
deansheather: This message should also mention that the workspace may not be started.

Tests.App/Services/UriHandlerTest.cs
```diff
@@ -0,0 +1,184 @@
...
+[TestFixture]
+public class UriHandlerTest
+{
+    [SetUp]
+    public void SetupMocksAndUriHandler()
+    {
+        Serilog.Log.Logger = new LoggerConfiguration().MinimumLevel.Debug().WriteTo.NUnitOutput().CreateLogger();
+        var builder = Host.CreateApplicationBuilder();
+        builder.Services.AddSerilog();
+        var logger = (ILogger<UriHandler>)builder.Build().Services.GetService(typeof(ILogger<UriHandler>))!;
```
deansheather, line 21: Creating an application builder for this seems excessive as it's only being used to create `ILogger` instances. Is there no way to just get an `ILogger` from serilog and pass that through without using application hosting?

spikecurtis: If there is, I haven't figured it out.

deansheather (approved):

spikecurtis:
### Merge activity

* **May 8, 5:54 AM EDT**: @spikecurtis merged this pull request with [Graphite](https://app.graphite.dev/github/pr/coder/coder-desktop-windows/87).

Issue #76, opened 2025-04-23
Fix Coder icon missing on forms


ibetitsmike:
Create a new base class containing the Coder icon and propagate that throughout the app to make sure all popups have that icon.

PR #89, opened 2025-05-06
feat: add coder icon to all forms

ibetitsmike:
Closes: #76


App/Controls/TitleBarIcon.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public static class TitleBarIcon
+    {
+        public static void SetTitlebarIcon(Window window)
+        {
+            var hwnd = WindowNative.GetWindowHandle(window);
+            var windowId = Win32Interop.GetWindowIdFromWindow(hwnd);
+            AppWindow.GetFromWindowId(windowId).SetIcon("coder.ico");
+        }
+    }
```
deansheather, line 18: Should go in `App/Utils` or something (which is a dir I'm using in #86) as it's not a `UserControl`

deansheather (approved):

App/Controls/TitleBarIcon.cs
ibetitsmike: Done

Issue #96, opened 2025-05-09
UI should show when 'Sign out' is disabled


ibetitsmike:
Currently Sign out button is disabled when Coder Connect is ON, however the fact that the button is disabled is not reflected in the UI.

PR #98, opened 2025-05-12
feat: gray out disabled Sign Out button

ibetitsmike:
Closes: #96


deansheather (commented):
Could you make this into a UserControl so we can reuse it? I imagine a UserControl that accepts only one UIElement that can be disabled, and a tooltip.

ethanndickson:
Should the sign out button just stop Coder Connect and then sign out?

PR #101, opened 2025-05-13
fix: fix daemon.lock race on mutagen startup

spikecurtis:
I found the source of the issue where mutagen would fail to acquire the lock on `daemon.lock` at startup.

The MutagenClient attempts to lock the `daemon.lock` file while it is starting, so that it can fail fast if the daemon is not running.

While well meaning, this creates a race condition because as soon as we start the daemon process we create a MutagenClient so that we can talk to the daemon over its API. The MutagenClient might be holding the lock or have the lockfile open at the exact moment the daemon itself attempts to acquire and lock the file. The daemon immediately exits in that case and doesn't retry locking the file.

I've just removed the preflight checks on the `daemon.lock`, since we don't want Coder Desktop to ever mess with that file (outside of tests).

### DIFF
--- a/App/Services/MutagenController.cs
+++ b/App/Services/MutagenController.cs
@@ -441,6 +441,7 @@ public sealed class MutagenController : ISyncSessionController
     /// </summary>
     private async Task<MutagenClient> EnsureDaemon(CancellationToken ct)
     {
+        _logger.LogDebug("EnsureDaemon called");
         ObjectDisposedException.ThrowIf(_disposing, typeof(MutagenController));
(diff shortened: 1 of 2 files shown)


spikecurtis:
* **#101** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/101?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/101?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #102, opened 2025-05-14
chore: make hostname suffix mutable in views

spikecurtis:
Part 1 of #49

Makes the workspace suffix dynamic in the views. A later PR in this stack will fetch the suffix and apply it to any views if it changes.

If the suffix doesn't match the FQDN provided by the VPN service, we won't gray out anything when we show the name, but if we later get a suffix that matches, it will update.

PR #103, opened 2025-05-15
feat: fetch hostname suffix from API

### LINKED ISSUE #49: Incorporate dynamic workspace suffix into UI
Currently we're hardcoding the `.coder` suffix in the menu items.
With https://github.com/coder/coder/issues/16828 the `HostnameSuffix` will become a configurable value, thus we need to read that value and adjust the UI accordingly.

spikecurtis:
Fixes #49

Adds support to query the hostname suffix from Coder server, and then propagates any changes to the agent view models.

### DIFF
--- a/App/Services/HostnameSuffixGetter.cs
+++ b/App/Services/HostnameSuffixGetter.cs
@@ -78,8 +78,7 @@ public class HostnameSuffixGetter : IHostnameSuffixGetter
             _dirty = false;
         }
 
-        var client = _clientFactory.Create(credentials.CoderUrl!.ToString());
-        client.SetSessionToken(credentials.ApiToken!);
+        var client = _clientFactory.Create(credentials);
         using var timeoutSrc = new CancellationTokenSource(TimeSpan.FromSeconds(10));
         var connInfo = await client.GetAgentConnectionInfoGeneric(timeoutSrc.Token);
 
--- a/App/Services/HostnameSuffixGetter.cs
+++ b/App/Services/HostnameSuffixGetter.cs
@@ -1,0 +1,145 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Coder.Desktop.App.Models;
+using Coder.Desktop.CoderSdk.Coder;
+using Coder.Desktop.Vpn.Utilities;
+using Microsoft.Extensions.Logging;
+
+namespace Coder.Desktop.App.Services;
+
+public interface IHostnameSuffixGetter
+{
+    public event EventHandler<string> SuffixChanged;
+
+    public string GetCachedSuffix();
+}
+
+public class HostnameSuffixGetter : IHostnameSuffixGetter
+{
+    private const string DefaultSuffix = ".coder";
+
+    private readonly ICredentialManager _credentialManager;
+    private readonly ICoderApiClientFactory _clientFactory;
+    private readonly ILogger<HostnameSuffixGetter> _logger;
+
+    // _lock protects all private (non-readonly) values
+    private readonly RaiiSemaphoreSlim _lock = new(1, 1);
+    private string _domainSuffix = DefaultSuffix;
+    private bool _dirty = false;
+    private bool _getInProgress = false;
+    private CredentialModel _credentialModel = new() { State = CredentialState.Invalid };
+
+    public event EventHandler<string>? SuffixChanged;
+
+    public HostnameSuffixGetter(ICredentialManager credentialManager, ICoderApiClientFactory apiClientFactory,
+        ILogger<HostnameSuffixGetter> logger)
+    {
+        _credentialManager = credentialManager;
+        _clientFactory = apiClientFactory;
+        _logger = logger;
+        credentialManager.CredentialsChanged += HandleCredentialsChanged;
+        HandleCredentialsChanged(this, _credentialManager.GetCachedCredentials());
+    }
+
+    ~HostnameSuffixGetter()
+    {
+        _credentialManager.CredentialsChanged -= HandleCredentialsChanged;
+    }
+
+    private void HandleCredentialsChanged(object? sender, CredentialModel credentials)
+    {
+        using var _ = _lock.Lock();
+        _logger.LogDebug("credentials updated with state {state}", credentials.State);
+        _credentialModel = credentials;
+        if (credentials.State != CredentialState.Valid) return;
+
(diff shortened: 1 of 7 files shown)


spikecurtis:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/103?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#103** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/103?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/103?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#102** <a href="https://app.graphite.dev/github/pr/coder/coder-desktop-windows/102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

App/Services/HostnameSuffixGetter.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            _getInProgress = true;
+            Task.Run(Refresh).ContinueWith(MaybeRefreshAgain);
+        }
+    }
+
+    private async Task Refresh()
+    {
+        _logger.LogDebug("refreshing domain suffix");
+        CredentialModel credentials;
```
deansheather, line 68: You should probably make `CredentialModel` implement `CoderSdk.Coder.ICoderApiClientCredentialProvider`, then you can just call `_clientFactory.Create(credentials)` below

App/Services/HostnameSuffixGetter.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            _logger.LogError(prev.Exception, "failed to query domain suffix");
+            // back off here before retrying. We're just going to use a fixed, long
+            // delay since this just affects UI stuff; we're not in a huge rush as
+            // long as we eventually get the right value.
+            await Task.Delay(TimeSpan.FromSeconds(10));
+        }
+
+        using var l = await _lock.LockAsync(CancellationToken.None);
+        if ((_dirty || prev.IsFaulted) && _credentialModel.State == CredentialState.Valid)
```
deansheather, line 124: `_dirty` seems to only be true here if the credentials became invalid between when the event was handled and the task started. So the only way for this to trigger would be if credentials became valid, then became invalid, then became valid again, which seems very unlikely to actually happen. The user would need to sign in or start the app, then sign out, then enter their credentials and sign back in (with API auth check latency included).

You could probably also accomplish the same thing without needing a bool by throwing in the credential state guard at the top of `Refresh()`, and only relying on `prev.IsFaulted` here

App/Services/HostnameSuffixGetter.cs
```diff
@@ -0,0 +1,145 @@
...
+    {
+        _credentialManager.CredentialsChanged -= HandleCredentialsChanged;
+    }
+
+    private void HandleCredentialsChanged(object? sender, CredentialModel credentials)
+    {
+        using var _ = _lock.Lock();
+        _logger.LogDebug("credentials updated with state {state}", credentials.State);
+        _credentialModel = credentials;
+        if (credentials.State != CredentialState.Valid) return;
```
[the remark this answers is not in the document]
deansheather: Maybe it would be better to just always create the task if the credentials changed (i.e. remove this check here) and do the check once in the task.

App/Services/HostnameSuffixGetter.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        _dirty = true;
+        if (!_getInProgress)
+        {
+            _getInProgress = true;
+            Task.Run(Refresh).ContinueWith(MaybeRefreshAgain);
+        }
+    }
+
+    private async Task Refresh()
```
deansheather, line 65: This should probably just take the credentials model as an argument.

App/Services/HostnameSuffixGetter.cs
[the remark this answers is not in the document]
spikecurtis: I don't think we should set the stored domain back to the default. The most common scenario for getting invalid creds is probably a logout or expired token --- so it's not particularly likely that the Default is more correct.

In any case, once we sign back in we'll get the correct value.

--------------

In terms of avoiding the check and always creating the task, seems like a lot of churn (locking, creating tasks, dropping logs) to avoid one conditional.

App/Services/HostnameSuffixGetter.cs
spikecurtis: The idea of `_dirty` is that while we are making the network call a new, valid credential for a different deployment could come in. That is, just because we had a valid credential when we started our work, and have a valid credential at the end, it doesn't mean that _it was the same credential_. If that happens then we refresh to ensure we have the correct value from the new credential.

App/Services/HostnameSuffixGetter.cs
spikecurtis: No, we want this to operate on the "latest" credentials since we're lock anyway to set `_dirty`.

App/Services/HostnameSuffixGetter.cs
```diff
@@ -0,0 +1,145 @@
...
+    private readonly ICredentialManager _credentialManager;
+    private readonly ICoderApiClientFactory _clientFactory;
+    private readonly ILogger<HostnameSuffixGetter> _logger;
+
+    // _lock protects all private (non-readonly) values
+    private readonly RaiiSemaphoreSlim _lock = new(1, 1);
+    private string _domainSuffix = DefaultSuffix;
+    private bool _dirty = false;
+    private bool _getInProgress = false;
+    private CredentialModel _credentialModel = new() { State = CredentialState.Invalid };
```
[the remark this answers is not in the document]
spikecurtis: I tried this, but it complicates things considerably in terms of race conditions because you can't synchronize changes to `GetCachedCredentials` with the other items we track here like `_dirty`, so its hard to guarantee the correct behavior. Writing updates to this while holding the lock ends up being much simpler.

deansheather (approved):

PR #109, opened 2025-05-27
feat: enabled sign out and animated window resize

ibetitsmike:
Closes: #96

### DIFF
--- a/App/Services/RpcController.cs
+++ b/App/Services/RpcController.cs
@@ -234,6 +234,11 @@ public class RpcController : IRpcController
             MutateState(state => { state.VpnLifecycle = VpnLifecycle.Unknown; });
             throw new VpnLifecycleException($"Failed to stop VPN. Service reported failure: {reply.Stop.ErrorMessage}");
         }
+
+        if (reply.Stop.Success)
+        {
+            MutateState(state => { state.VpnLifecycle = VpnLifecycle.Stopped; });
+        }
     }
 
     public async ValueTask DisposeAsync()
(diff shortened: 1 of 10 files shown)


App/Services/RpcController.cs
```diff
@@ -234,6 +234,11 @@ public async Task StopVpn(CancellationToken ct = default)
             MutateState(state => { state.VpnLifecycle = VpnLifecycle.Unknown; });
             throw new VpnLifecycleException($"Failed to stop VPN. Service reported failure: {reply.Stop.ErrorMessage}");
         }
+
+        if (reply.Stop.Success)
+        {
+            MutateState(state => { state.VpnLifecycle = VpnLifecycle.Stopped; });
+        }
```
deansheather, line 241:
```suggestion
        }

        MutateState(state => { state.VpnLifecycle = VpnLifecycle.Stopped; });
```

App/Views/TrayWindow.xaml.cs
```diff
@@ -140,7 +174,27 @@ public void SetRootFrame(Page page)
...
     {
-        MoveAndResize(e.NewSize.Height);
+        AnimateWindowHeight(e.NewSize.Height);
+    }
+
+    private void AnimateWindowHeight(double targetHeight)
+    {
+        // Remember where we start
+        _lastWindowHeight = AppWindow.Size.Height;
+        _resizeInProgress = true;
```
deansheather, line 184: This never gets set back to false AFAIK

deansheather (approved):

App/Views/TrayWindow.xaml.cs
ibetitsmike: It was a leftover I forgot to clean-up. Initial implementation was using a semaphore to order transactions, but I saw BeginTime allows you to delay enough to order them. I added Skip when user clicks quickly and that resolves all issues. 

Added comments for better understanding.

Issue #38, opened 2025-03-01
Windows installer doesn't have an icon while running


deansheather:
The .exe itself has an icon but the installer window does not have one in the window titlebar or the taskbar.

I'm guessing this is because the embedded `wixstdba.exe` needs to be modified to use the icon or something

![Image](https://github.com/user-attachments/assets/058b68eb-6d9a-4929-8630-d0fd61b4e75d)

![Image](https://github.com/user-attachments/assets/8368bb02-1481-41fb-9de9-ad73331cb342)

matifali:
@ibetitsmike I feel we have already done it, if I am not mistaken, and it's something different.

ibetitsmike:
That's a bug in Wix5 toolset: https://github.com/wixtoolset/issues/issues/8104

PR #112, opened 2025-05-29
feat: added icon to installer's bootstrap application

ibetitsmike:
Closes: #38 

WiX 5 introduced a bug that stopped respecting the icon file on the theme's Window. This is described in this issue: https://github.com/wixtoolset/issues/issues/8104

This PR introduces:

- an additional option in our `build-bootstrapper` to allow optional setting of the `theme xml`, 
- adds a copy of the `RtfLargeTheme.xml` from WiX `5.0.2` with the added attribute of `IconFile`

image

### DIFF
--- a/Installer/Program.cs
+++ b/Installer/Program.cs
@@ -423,7 +423,7 @@ public class Program
         if (opts.ThemeXmlPath != null)
         {
             bundle.Application.ThemeFile = opts.ThemeXmlPath;
-            bundle.Application.Payloads = 
+            bundle.Application.Payloads =
             [
                 new ExePackagePayload
                 {
--- a/Installer/Program.cs
+++ b/Installer/Program.cs
@@ -119,6 +119,9 @@ public class BootstrapperOptions : SharedOptions
     [Option('w', "windows-app-sdk-path", Required = true, HelpText = "Path to the Windows App Sdk package to embed")]
     public string WindowsAppSdkPath { get; set; }
 
+    [Option('t', "theme-xml-path", Required = false, HelpText = "Path to the theme .xml file to use for the installer")]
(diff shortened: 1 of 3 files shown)


Installer/Program.cs
```diff
@@ -119,6 +119,9 @@ public class BootstrapperOptions : SharedOptions
     [Option('w', "windows-app-sdk-path", Required = true, HelpText = "Path to the Windows App Sdk package to embed")]
     public string WindowsAppSdkPath { get; set; }
 
+    [Option('t', "theme-xml-path", Required = false, HelpText = "Path to the theme .xml file to use for the installer")]
+    public string ThemeXmlPath { get; set; }
```
deansheather, line 123: Should be `string?` if it's nullable

scripts/files/RtfThemeLarge.xml
```diff
@@ -0,0 +1,119 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information. -->
+<!-- Downloaded from https://github.com/wixtoolset/wix/blob/v5.0.2/src/ext/Bal/stdbas/Resources/RtfLargeTheme.xml -->
```
deansheather, line 3: We likely need to include the license text. We can probably do it as a comment in this file to avoid writing another `LICENSE.xyz.TXT` file in the repo.

deansheather, line 3: Can you add a comment explaining why we downloaded this and what you changed? Just in case we need to do it again. I'd like to the issue

deansheather (approved):

PR #113, opened 2025-05-29
added new settings dialog + settings manager

ibetitsmike:


App/Services/SettingsManager.cs
```diff
@@ -0,0 +1,113 @@
...
+            // Corrupted file – start fresh.
+            return new();
+        }
+    }
+
+    private void Persist()
+    {
+        using var fs = File.Create(_settingsFilePath);
+        var options = new JsonSerializerOptions { WriteIndented = true };
+        JsonSerializer.Serialize(fs, _cache, options);
```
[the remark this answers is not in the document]
ibetitsmike: Tbh it's a rare use case that doesn't break the flow - user shouldn't edit their settings in runtime and expect the settings to persist. However, it's a pretty simple fix (although requires some magic operating the stream position to ensure we correctly override the file).

App/App.xaml.cs
```diff
@@ -109,6 +120,8 @@ public App()
         _services = services.BuildServiceProvider();
         _logger = (ILogger<App>)_services.GetService(typeof(ILogger<App>))!;
         _uriHandler = (IUriHandler)_services.GetService(typeof(IUriHandler))!;
+        _settingsManager = (ISettingsManager)_services.GetService(typeof(ISettingsManager))!;
+        _appLifetime = (IHostApplicationLifetime)_services.GetRequiredService<IHostApplicationLifetime>();
```
deansheather, line 124:
```suggestion
        _logger = _services.GetRequiredService<ILogger<App>>();
        _uriHandler = _services.GetRequiredService<IUriHandler>();
        _settingsManager = _services.GetRequiredService<ISettingsManager>();
        _appLifetime = _services.GetRequiredService<IHostApplicationLifetime>();
```

App/App.xaml.cs
```diff
@@ -192,6 +155,63 @@ protected override void OnLaunched(LaunchActivatedEventArgs args)
...
+        {
+            if (loadCredsTask.IsFaulted)
+                _logger.LogError(loadCredsTask.Exception!.GetBaseException(),
+                                 "Failed to load credentials");
+
+            if (reconnectTask.IsFaulted)
+                _logger.LogError(reconnectTask.Exception!.GetBaseException(),
+                                 "Failed to connect to VPN service");
+
+            return;
```
deansheather, line 187: We still should attempt to start file sync or it will never initialize

App/App.xaml.cs
```diff
@@ -192,6 +155,63 @@ protected override void OnLaunched(LaunchActivatedEventArgs args)
         };
     }
 
+    /// <summary>
+    /// Loads stored VPN credentials, reconnects the RPC controller,
+    /// and (optionally) starts the VPN tunnel on application launch.
+    /// </summary>
+    private async Task InitializeServicesAsync(CancellationToken cancellationToken = default)
```
deansheather, line 162: This pattern is already looking a lot better!

App/Services/SettingsManager.cs
```diff
@@ -0,0 +1,189 @@
...
+    /// <summary>
+    /// Returns the value of the StartOnLogin setting. Returns <c>false</c> if the key is not found.
+    /// </summary>
+    bool StartOnLogin { get; set; }
+
+    /// <summary>
+    /// Returns the value of the ConnectOnLaunch setting. Returns <c>false</c> if the key is not found.
+    /// </summary>
+    bool ConnectOnLaunch { get; set; }
+}
```
deansheather, line 23: I still think this should just become something like:

```cs
public interface ISettingsManager<T> {
  public Task<T> Read(CancellationToken ct = default); // may return from cache if unchanged, always returns a clone
  public Task Write(T settings, CancellationToken ct = default);
}
```

The setters you currently have use synchronous I/O, which is not something I think you can really work around without having a dedicated method for saving values instead.

App/Services/SettingsManager.cs
```diff
@@ -0,0 +1,189 @@
...
+    private readonly string _settingsFilePath;
+    private Settings _settings;
+    private readonly string _fileName = "app-settings.json";
+    private readonly string _appName = "CoderDesktop";
+    private readonly object _lock = new();
+
+    public const string ConnectOnLaunchKey = "ConnectOnLaunch";
+    public const string StartOnLoginKey = "StartOnLogin";
+
+    public bool StartOnLogin
```
deansheather, line 40: Since the registry is the source of truth for this, it should probably just be removed from the settings file entirely.

App/Services/StartupManager.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    public bool Enable()
+    {
+        if (IsDisabledByPolicy())
+            return false;
+
+        string exe = Process.GetCurrentProcess().MainModule!.FileName;
+        try
+        {
+            using var key = Registry.CurrentUser.OpenSubKey(RunKey, writable: true)
```
deansheather, line 48: Is there any (easy) way we can clean this up on uninstall?

App/Views/Pages/SettingsMainPage.xaml
```diff
@@ -0,0 +1,59 @@
...
+                                       Header="Connect on launch"
+                                       HeaderIcon="{ui:FontIcon Glyph=&#xE8AF;}"
+                                       >
+                    <ToggleSwitch IsOn="{x:Bind ViewModel.ConnectOnLaunch, Mode=TwoWay}" />
+                </controls:SettingsCard>
+
+            </StackPanel>
+        </Grid>
+    </ScrollViewer>
+
```
deansheather, line 58:
```suggestion
```

App/Views/Pages/SettingsMainPage.xaml
```diff
@@ -0,0 +1,59 @@
...
+                </controls:SettingsCard>
+
+                <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="Coder Connect" />
+                <controls:SettingsCard Description="This setting controls whether Coder Connect automatically starts with Coder Desktop. "
+                                       Header="Connect on launch"
+                                       HeaderIcon="{ui:FontIcon Glyph=&#xE8AF;}"
+                                       >
+                    <ToggleSwitch IsOn="{x:Bind ViewModel.ConnectOnLaunch, Mode=TwoWay}" />
+                </controls:SettingsCard>
+
```
deansheather, line 54:
```suggestion
```

App/App.xaml.cs
ibetitsmike: Isn't that done in lines 203:208 in this file? just below this one

deansheather: It seems like this return here would prevent that, unless I'm mistaken. We'd want to always initialize file sync even if RPC or credentials failed to load.

App/Services/StartupManager.cs
ibetitsmike: We can, but this will also basically clear the setting every team you update (I believe you run the uninstall then too, right)?

ibetitsmike: Follow up here: #121

App/Services/SettingsManager.cs
```diff
@@ -1,65 +1,57 @@
...
+    /// <param name="ct"></param>
+    /// <returns></returns>
+    public Task Write(T settings, CancellationToken ct = default);
     /// <summary>
-    /// Returns the value of the ConnectOnLaunch setting. Returns <c>false</c> if the key is not found.
+    /// Returns null if the settings are not cached or not available.
     /// </summary>
-    bool ConnectOnLaunch { get; set; }
+    /// <returns></returns>
+    public T? GetFromCache();
```
deansheather, line 37: Rather than have this, `Read` should probably return from cache by default. I know this goes against a point I made in a previous review but if we're going with the Read/Write API then it should just be all in

deansheather: Also, Read should probably not fail if it failed to read/parse the file. IDK what good behavior is, but I think for now just returning a default config is OK

App/Services/SettingsManager.cs
```diff
@@ -81,109 +73,129 @@ public SettingsManager(string? settingsFilePath = null)
...
+{
+    /// <summary>
+    /// CoderConnect settings version. Increment this when the settings schema changes.
     /// In future iterations we will be able to handle migrations when the user has
     /// an older version.
     /// </summary>
     public int Version { get; set; }
-    public Dictionary<string, JsonElement> Options { get; set; }
+    public bool ConnectOnLaunch { get; set; }
+    public static string SettingsFileName { get; } = "coder-connect-settings.json";
```
deansheather, line 180: Probs should be at the top of the class

App/ViewModels/SettingsViewModel.cs
```diff
@@ -20,16 +20,24 @@ public partial class SettingsViewModel : ObservableObject
...
         _logger = logger;
-        ConnectOnLaunch = _settingsManager.ConnectOnLaunch;
-        StartOnLogin = _settingsManager.StartOnLogin;
+        // Application settings are loaded on application startup,
+        // so we expect the settings to be available immediately.
+        var settingsCache = settingsManager.GetFromCache();
+        if (settingsCache is not null)
+        {
+            _connectSettings = settingsCache.Clone();
+        }
```
deansheather, line 38: When returning a cached value we should probably just always clone it from within ISettingsManager

deansheather (approved):

Issue #119, opened 2025-06-06
Improve the logic for drawing the tray window


ibetitsmike:
Currently the tray window blindly believes that the cursor position is where to draw the Tray Windows. 
We can calculate where to draw the window based on the screen size.

PR #125, opened 2025-06-11
feat: calculate taskbar position and display on top of it

ibetitsmike:
Closes: #119

### DIFF
--- a/App/Views/TrayWindow.xaml.cs
+++ b/App/Views/TrayWindow.xaml.cs
@@ -33,8 +33,6 @@ public sealed partial class TrayWindow : Window
     private int _lastWindowHeight;
     private Storyboard? _currentSb;
 
-    private NativeApi.POINT? _lastActivatePosition;
-
     private readonly IRpcController _rpcController;
     private readonly ICredentialManager _credentialManager;
     private readonly ISyncSessionController _syncSessionController;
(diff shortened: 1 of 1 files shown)


App/Views/TrayWindow.xaml.cs
```diff
@@ -257,41 +239,38 @@ private SizeInt32 CalculateWindowSize(double height)
...
+        var tb = GetTaskbarInfo(area);
 
-        var workArea = DisplayArea.GetFromPoint(
-            new PointInt32(cursorPosition.Value.X, cursorPosition.Value.Y),
-            DisplayAreaFallback.Primary
-        ).WorkArea;
-
-        // Adjust if the window goes off the right edge of the display.
-        if (x + width > workArea.X + workArea.Width) x = workArea.X + workArea.Width - width;
+        int x, y;
```
deansheather, line 250: The previous calculations had some hardcoded plus or minus `10` as a padding, shouldn't we do something similar to avoid it directly touching the taskbar or the side of the screen?

App/Views/TrayWindow.xaml.cs
```diff
@@ -257,41 +239,38 @@ private SizeInt32 CalculateWindowSize(double height)
...
+                x = workArea.X + workArea.Width - panelSize.Width;
+                y = workArea.Y + workArea.Height - panelSize.Height - tb.Gap;
+                break;
 
-        // Adjust if the window goes off the top edge of the display (somehow).
-        if (y < workArea.Y) y = workArea.Y;
+            default: // right or bottom when not auto-hiding
+                x = workArea.X + workArea.Width - panelSize.Width;
+                y = bounds.Y + bounds.Height - panelSize.Height - tb.Gap;
+                break;
```
deansheather, line 272: I don't get why we wouldn't just always use `workArea.Height` here, the comment doesn't do a very good job explaining why we wouldn't prefer to always use the workArea.Height.

App/Views/TrayWindow.xaml.cs
```diff
@@ -338,4 +317,70 @@ public struct POINT
             public int Y;
         }
     }
```
deansheather, line 319:
```suggestion
    }

```

App/Views/TrayWindow.xaml.cs
ibetitsmike: We're already getting a margin because the panel inside the root frame is a bit wider than the actual window. Adding artificial padding is making the window look weird.

deansheather (approved):

Issue #106, opened 2025-05-22
Show latency on workspace view


ibetitsmike:
Related to: https://github.com/coder/coder-desktop-macos/issues/64 (same implementation)

PR #134, opened 2025-06-23
chore: added latency tooltips on workspaces

ibetitsmike:
Fixes: #106 

image
image

### DIFF
--- a/App/ViewModels/AgentViewModel.cs
+++ b/App/ViewModels/AgentViewModel.cs
@@ -242,17 +242,17 @@ public partial class AgentViewModel : ObservableObject, IModelUpdateable<AgentVi
                 """
                 );
             }
-            else if (PreferredDerpLatency != null)
+            else if (Latency != null)
             {
                 description.Append($"""
                 You're connected through a DERP relay. {highLatencyWarning}
                 We'll switch over to peer-to-peer when available.
 
-                Total latency: {PreferredDerpLatency.Value.Milliseconds} ms
+                Total latency: {Latency.Value.Milliseconds} ms
                 """
                 );
 
-                if (PreferredDerp != null && Latency != null)
+                if (PreferredDerpLatency != null)
                 {
                     description.Append($"\nYou ↔ {PreferredDerp}: {PreferredDerpLatency.Value.Milliseconds} ms");
 


App/ViewModels/AgentViewModel.cs
```diff
@@ -182,6 +191,75 @@ public string FullyQualifiedDomainName
...
+
+                    // Guard against negative values if the two readings were taken at different times
+                    if (derpToWorkspaceEstimatedLatency > TimeSpan.Zero)
+                    {
+                        description.Append($"\n{PreferredDerp} ms ↔ {WorkspaceName}: {derpToWorkspaceEstimatedLatency.Value.Milliseconds} ms");
+                    }
+                }
+            }
+            if (LastHandshake != null)
+                description.Append($"\n\nLast handshake: {LastHandshake?.ToString() ?? "Unknown"}");
```
deansheather, line 251:
```suggestion
                description.Append($"\n\nLast handshake: {LastHandshake.ToString()}");
```

App/ViewModels/TrayWindowViewModel.cs
```diff
@@ -222,12 +224,26 @@ private void UpdateFromRpcModel(RpcModel rpcModel)
...
 
+
             var lastHandshakeAgo = DateTime.UtcNow.Subtract(agent.LastHandshake.ToDateTime());
-            var connectionStatus = lastHandshakeAgo < TimeSpan.FromMinutes(5)
-                ? AgentConnectionStatus.Green
-                : AgentConnectionStatus.Yellow;
+
+            // For compatibility with older deployments, we assume that if the
+            // last ping is null, the agent is healthy.
+            var isLatencyAcceptable = agent.LastPing != null ? agent.LastPing.Latency.ToTimeSpan() < HealthyPingThreshold : true;
```
deansheather, line 232: Did you check that the behavior of the C# protobuf library does this? They usually just use a zero value when it's not set to be consistent with other protobuf implementations

App/ViewModels/AgentViewModel.cs
```diff
@@ -182,6 +191,75 @@ public string FullyQualifiedDomainName
...
+
+            var tooltip = description.ToString().TrimEnd('\n', ' ');
+
+            if (string.IsNullOrEmpty(tooltip))
+                return "No connection information available.";
```
deansheather, line 256: Could this just be an `else` on the `if (p2p) {} else if (derp) {}` at the top?

App/ViewModels/AgentViewModel.cs
```diff
@@ -73,10 +82,10 @@ public AgentViewModel CreateDummy(IAgentExpanderHost expanderHost, Uuid id,
...
 public enum AgentConnectionStatus
 {
-    Green,
-    Yellow,
-    Red,
-    Gray,
+    Healthy,
+    Unhealthy,
+    NoRecentHandshake,
+    Offline,
```
ethanndickson, line 88: On mac, I also have a yellow `connecting` status when the `LastHandshake` is missing or negative, otherwise the status dot goes immediately red, and I think that's misleading/needlessly scary.

App/ViewModels/AgentViewModel.cs
```diff
@@ -182,6 +191,75 @@ public string FullyQualifiedDomainName
...
+                        description.Append($"\n{PreferredDerp} ms ↔ {WorkspaceName}: {derpToWorkspaceEstimatedLatency.Value.Milliseconds} ms");
+                    }
+                }
+            }
+            if (LastHandshake != null)
+                description.Append($"\n\nLast handshake: {LastHandshake?.ToString() ?? "Unknown"}");
+
+            var tooltip = description.ToString().TrimEnd('\n', ' ');
+
+            if (string.IsNullOrEmpty(tooltip))
```
ethanndickson, line 255: Is this necessary? When would this happen?

I'm not handling this on mac, if it can happen:
```swift
    var statusString: String {
        switch status {
        case .okay, .high_latency:
            break
        default:
            return status.description
        }

        guard let lastPing else {
            // Either:
            // - Old coder deployment
            // - We haven't received any pings yet
            return status.description
        }

        let highLatencyWarning = status == .high_latency ? "(High latency)" : ""

        var str: String
        if lastPing.didP2p {
            str = """
            You're connected peer-to-peer. \(highLatencyWarning)

            You ↔ \(lastPing.latency.prettyPrintMs) ↔ \(wsName)
            """
        } else {
            str = """
            You're connected through a DERP relay. \(highLatencyWarning)
            We'll switch over to peer-to-peer when available.

            Total latency: \(lastPing.latency.prettyPrintMs)
            """
            // We're not guranteed to have the preferred DERP latency
            if let preferredDerpLatency = lastPing.preferredDerpLatency {
                str += "\nYou ↔ \(lastPing.preferredDerp): \(preferredDerpLatency.prettyPrintMs)"
                let derpToWorkspaceEstLatency = lastPing.latency - preferredDerpLatency
                // We're not guaranteed the preferred derp latency is less than
                // the total, as they might have been recorded at slightly
                // different times, and we don't want to show a negative value.
                if derpToWorkspaceEstLatency > 0 {
                    str += "\n\(lastPing.preferredDerp) ↔ \(wsName): \(derpToWorkspaceEstLatency.prettyPrintMs)"
                }
            }
        }
        str += "\n\nLast handshake: \(lastHandshake?.relativeTimeString ?? "Unknown")"
        return str
    }
    ```

File: App/ViewModels/TrayWindowViewModel.cs (hunk 2 above)
Comment (ibetitsmike) [reply App/ViewModels/TrayWindowViewModel.cs:232]: Yes, for reference types it will be null.

File: App/ViewModels/AgentViewModel.cs (hunk 5 above)
Comment (ibetitsmike) [reply App/ViewModels/AgentViewModel.cs]: I believe you handle it here:
` 
guard let lastPing else {
            // Either:
            // - Old coder deployment
            // - We haven't received any pings yet
            return status.description
        }
`

I have now replicated this logic by printing the status.

File: App/ViewModels/TrayWindowViewModel.cs [hunk 6]
```diff
@@ -222,10 +224,28 @@ private void UpdateFromRpcModel(RpcModel rpcModel)
             if (string.IsNullOrWhiteSpace(fqdn))
                 continue;
 
-            var lastHandshakeAgo = DateTime.UtcNow.Subtract(agent.LastHandshake.ToDateTime());
-            var connectionStatus = lastHandshakeAgo < TimeSpan.FromMinutes(5)
-                ? AgentConnectionStatus.Green
-                : AgentConnectionStatus.Yellow;
+            var connectionStatus = AgentConnectionStatus.Healthy;
+
+            if (agent.LastHandshake != null && agent.LastHandshake.ToDateTime() != default && agent.LastHandshake.ToDateTime() < DateTime.UtcNow)
```
ethanndickson, line 229: 👍

App/ViewModels/AgentViewModel.cs
```diff
@@ -182,6 +206,77 @@ public string FullyQualifiedDomainName
...
+                        description.Append($"\n{PreferredDerp} ms ↔ {WorkspaceName}: {derpToWorkspaceEstimatedLatency.Value.Milliseconds} ms");
+                    }
+                }
+            }
+            else
+            {
+                description.Append(ConnectionStatus.ToDisplayString());
+            }
+            if (LastHandshake != null)
+                description.Append($"\n\nLast handshake: {LastHandshake?.ToString()}");
```
ethanndickson, line 273: I have a relative timestamp (i.e. `5 seconds ago`) here but I actually gotta get rid of it, so this is fine.
( Have to get rid of it cause it only gets rendered once, and doesn't update on tooltip re-hovers 😦 )

ethanndickson (approved):

App/ViewModels/AgentViewModel.cs
ibetitsmike: Yeah, I check for Default and set to null in the viewmodel cause showing a made up handshake just felt wrong.

PR #140, opened 2025-06-26
chore: add connect/disconnect notifications on Coder Connect

ibetitsmike:
Fixes: #40

### DIFF
--- a/App/App.xaml.cs
+++ b/App/App.xaml.cs
@@ -27,7 +27,7 @@ using LaunchActivatedEventArgs = Microsoft.UI.Xaml.LaunchActivatedEventArgs;
 
 namespace Coder.Desktop.App;
 
-public partial class App : Application, IDispatcherQueueManager
+public partial class App : Application, IDispatcherQueueManager, INotificationHandler
 {
     private const string MutagenControllerConfigSection = "MutagenController";
     private const string UpdaterConfigSection = "Updater";
@@ -91,6 +91,7 @@ public partial class App : Application, IDispatcherQueueManager
         services.AddSingleton<IAgentApiClientFactory, AgentApiClientFactory>();
 
         services.AddSingleton<IDispatcherQueueManager>(_ => this);
+        services.AddSingleton<INotificationHandler>(_ => this);
         services.AddSingleton<ICredentialBackend>(_ =>
             new WindowsCredentialBackend(WindowsCredentialBackend.CoderCredentialsTargetName));
         services.AddSingleton<ICredentialManager, CredentialManager>();
@@ -334,5 +335,14 @@ public partial class App : Application, IDispatcherQueueManager
             return;
         }
         dispatcherQueue.TryEnqueue(action);
+    }
+
+    public void HandleNotificationActivation(IDictionary<string, string> args)
+    {
+        var app = (App)Current;
+        if (app != null && app.TrayWindow != null)
+        {
+            app.TrayWindow.Tray_Open();
+        }
     }
 }
--- a/App/Services/UserNotifier.cs
+++ b/App/Services/UserNotifier.cs
@@ -21,12 +21,23 @@ public interface IUserNotifier : INotificationHandler, IAsyncDisposable
     public void UnregisterHandler(string name);
 
     public Task ShowErrorNotification(string title, string message, CancellationToken ct = default);
+    /// <summary>
+    /// This method allows to display a Windows-native notification with an action defined in
+    /// <paramref name="handlerName"/> and provided <paramref name="args"/>.
+    /// </summary>
+    /// <param name="title">Title of the notification.</param>
+    /// <param name="message">Message to be displayed in the notification body.</param>
+    /// <param name="handlerName">Handler should be e.g. <c>nameof(Handler)</c> where <c>Handler</c>
+    /// implements <see cref="Coder.Desktop.App.Services.INotificationHandler" />.
+    /// If handler is <c>null</c> the action will open Coder Desktop.</param>
+    /// <param name="args">Arguments to be provided to the handler when executing the action.</param>
     public Task ShowActionNotification(string title, string message, string? handlerName, IDictionary<string, string>? args = null, CancellationToken ct = default);
 }
 
 public class UserNotifier : IUserNotifier
 {
     private const string CoderNotificationHandler = "CoderNotificationHandler";
+    private const string DefaultNotificationHandler = "DefaultNotificationHandler";
 
     private readonly AppNotificationManager _notificationManager = AppNotificationManager.Default;
     private readonly ILogger<UserNotifier> _logger;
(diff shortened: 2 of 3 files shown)


App/Services/UserNotifier.cs
```diff
@@ -90,20 +108,32 @@ public void HandleNotificationActivation(IDictionary<string, string> args)
...
             catch (Exception ex)
             {
-                logger.LogWarning(ex, "could not handle activation for notification with handler '{HandlerName}", handlerName);
+                _logger.LogWarning(ex, "could not handle activation for notification with handler '{HandlerName}", handlerName);
             }
         });
     }
 }
+
+public class DefaultNotificationHandler : INotificationHandler
```
deansheather, line 129: Can you make App implement this perhaps? Then it could just get registered like IDispatcherQueueManager does

App/Services/UserNotifier.cs
```diff
@@ -61,10 +71,18 @@ public Task ShowErrorNotification(string title, string message, CancellationToke
...
+        if (handlerName == null)
+        {
+            // Use default handler if no handler name is provided
+            handlerName = nameof(DefaultNotificationHandler);
+        }
+        else
+        {
+            if (!Handlers.TryGetValue(handlerName, out _))
+                throw new InvalidOperationException($"No action handler with the name '{handlerName}' is registered. Use null for default");
+        }
```
deansheather, line 85: Doesn't really need to be an `else` right? Still doesn't hurt to check first

App/Services/UserNotifier.cs
```diff
@@ -61,10 +71,18 @@ public Task ShowErrorNotification(string title, string message, CancellationToke
...
-            throw new InvalidOperationException($"No action handler with the name '{handlerName}' is registered.");
+        if (handlerName == null)
+        {
+            // Use default handler if no handler name is provided
+            handlerName = nameof(DefaultNotificationHandler);
+        }
+        else
+        {
+            if (!Handlers.TryGetValue(handlerName, out _))
+                throw new InvalidOperationException($"No action handler with the name '{handlerName}' is registered. Use null for default");
```
deansheather, line 84: This should be a comment on the interface method instead of in the exception description

App/Services/UserNotifier.cs
```diff
@@ -20,17 +21,26 @@ public interface IUserNotifier : INotificationHandler, IAsyncDisposable
...
+    private readonly ILogger<UserNotifier> _logger;
+    private readonly IDispatcherQueueManager _dispatcherQueueManager;
 
     private ConcurrentDictionary<string, INotificationHandler> Handlers { get; } = new();
 
+    public UserNotifier(ILogger<UserNotifier> logger, IDispatcherQueueManager dispatcherQueueManager)
+    {
+        _logger = logger;
+        _dispatcherQueueManager = dispatcherQueueManager;
+        Handlers.TryAdd(nameof(DefaultNotificationHandler), new DefaultNotificationHandler());
```
deansheather, line 41: This should probably fail rather than silently do nothing

deansheather, line 41: Probably should also prevent unregistering this handler string

App/Views/TrayWindow.xaml.cs
```diff
@@ -142,9 +151,54 @@ private void SetPageByState(RpcModel rpcModel, CredentialModel credentialModel,
         }
     }
 
+    private void NotifyUser(RpcModel rpcModel)
```
deansheather, line 154: `MaybeNotifyUser` cuz it doesn't always send one

App/Views/TrayWindow.xaml.cs
```diff
@@ -142,9 +151,54 @@ private void SetPageByState(RpcModel rpcModel, CredentialModel credentialModel,
...
+        prevVpnLifecycle = rpcModel.VpnLifecycle;
+
+        if (_aw.IsVisible)
+        {
+            return; // No need to notify if the window is not visible.
+        }
+
+        // Trigger notification
+        _userNotifier.ShowActionNotification(message, string.Empty, null, null, CancellationToken.None);
+    }
```
deansheather, line 196: I think this could be a lot cleaner with a method like:

```cs
var oldRpcLifeycle = prevRpcLifecycle;
var oldVpnLifecycle = this.prevVpnLifecycle;
prevRpcLifecycle = rpcModel.RpcLifecycle;
prevVpnLifecycle = rpcModel.VpnLifecycle;

var messages = new List<string>();

if (oldRpcLifecycle != RpcLifecycle.Disconnected && rpcModel.RpcLifecycle == RpcLifecycle.Disconnected)
{
  messages.Add("Disconnected from Coder background service.");
}

if (oldVpnLifecycle != rpcModel.VpnLifecycle)
{
  switch (rpcModel.VpnLifecycle) {
    case VpnLifecycle.Started:
      messages.Add("Coder Connect started.");
      break;
    case VpnLifecycle.Stopped:
      messages.Add("Coder Connect stopped.");
      break;
  }
}

if (messages.Count == 0) return;
if (_aw.IsVisible) return;

var message = string.Join(" ", messages);
_userNotifier.ShowActionNotification(message, string.Empty, null, null, CancellationToken.None);

File: App/Views/TrayWindow.xaml.cs [hunk 7]
```diff
@@ -297,7 +351,7 @@ private void Window_Activated(object sender, WindowActivatedEventArgs e)
     }
 
     [RelayCommand]
-    private void Tray_Open()
+    public void Tray_Open()
```
deansheather, line 354: Did this fix something?

ibetitsmike: needed to expose as public to make sure we can all it from another class

App/App.xaml.cs
```diff
@@ -335,4 +336,13 @@ public void RunInUiThread(DispatcherQueueHandler action)
         }
         dispatcherQueue.TryEnqueue(action);
     }
+
+    public void HandleNotificationActivation(IDictionary<string, string> args)
+    {
+        var app = (App)Current;
+        if (app != null && app.TrayWindow != null)
```
deansheather, line 343: Shouldn't this just be `if (TrayWindow != null) TrayWindow.Tray_Open();`?

App/Services/UserNotifier.cs
```diff
@@ -20,17 +21,40 @@ public interface IUserNotifier : INotificationHandler, IAsyncDisposable
...
+    private const string DefaultNotificationHandler = "DefaultNotificationHandler";
 
     private readonly AppNotificationManager _notificationManager = AppNotificationManager.Default;
+    private readonly ILogger<UserNotifier> _logger;
+    private readonly IDispatcherQueueManager _dispatcherQueueManager;
 
     private ConcurrentDictionary<string, INotificationHandler> Handlers { get; } = new();
 
+    public UserNotifier(ILogger<UserNotifier> logger, IDispatcherQueueManager dispatcherQueueManager,
+        INotificationHandler notificationHandler)
```
deansheather, line 49: Perhaps we should wrap it with a marker type like `public interface IDefaultNotificationHandler : INotificationHandler` so we don't inadvertently add another implementation that could get DI'd here, and only have the `App` implement `IDefaultNotificationHandler`

App/Views/TrayWindow.xaml.cs
```diff
@@ -142,9 +152,54 @@ private void SetPageByState(RpcModel rpcModel, CredentialModel credentialModel,
...
+        // This method is called when the state changes, but we don't want to notify
+        // the user if the state hasn't changed.
+        var isRpcLifecycleChanged = rpcModel.RpcLifecycle == RpcLifecycle.Disconnected && curRpcLifecycle != rpcModel.RpcLifecycle;
+        var isVpnLifecycleChanged = (rpcModel.VpnLifecycle == VpnLifecycle.Started || rpcModel.VpnLifecycle == VpnLifecycle.Stopped) && curVpnLifecycle != rpcModel.VpnLifecycle;
+
+        if (!isRpcLifecycleChanged && !isVpnLifecycleChanged)
+        {
+            return;
+        }
+
```
deansheather, line 166: I don't think you need any of these lines since they're essentially covered by what's below right?

deansheather (approved):

App/Views/TrayWindow.xaml.cs
ibetitsmike: Unfortunately no, I ignore the intermediate states of the Vpn lifecycle on purpose (and I don't update the prevRpcLifecycle for that reason).

PR #148, opened 2025-07-21
fix: opt-in tailscale vpn loop prevention

ibetitsmike:


App/Models/Settings.cs
```diff
@@ -34,6 +34,11 @@ public class CoderConnectSettings : ISettings<CoderConnectSettings>
     /// </summary>
     public bool ConnectOnLaunch { get; set; }
 
+    /// <summary>
+    /// When this is true Coder Connect will not attempt to protect against Tailscale loopback issues.
+    /// </summary>
+    public bool DisableTailscaleLoopProtection { get; set; }
```
deansheather, line 40: Maybe `EnableCorporateVpnSupport`

App/Views/Pages/SettingsMainPage.xaml
```diff
@@ -44,6 +44,12 @@
                                        >
                     <ToggleSwitch IsOn="{x:Bind ViewModel.ConnectOnLaunch, Mode=TwoWay}" />
                 </controls:SettingsCard>
+                <controls:SettingsCard Description="This setting controls whether Coder Connect should bypass VPN loop preventation mechanism. Turn this ON only when dealing with a Coder deployment behind a corporate VPN."
+                       Header="Disable corporate VPN loop protection"
```
deansheather, line 48:
```suggestion
                       Header="Enable support for corporate VPNs"
```

App/Views/Pages/SettingsMainPage.xaml
```diff
@@ -44,6 +44,12 @@
                                        >
                     <ToggleSwitch IsOn="{x:Bind ViewModel.ConnectOnLaunch, Mode=TwoWay}" />
                 </controls:SettingsCard>
+                <controls:SettingsCard Description="This setting controls whether Coder Connect should bypass VPN loop preventation mechanism. Turn this ON only when dealing with a Coder deployment behind a corporate VPN."
```
deansheather, line 47:
```suggestion
                <controls:SettingsCard Description="This setting loosens some VPN loop protection checks in Coder Connect, allowing traffic to flow to a Coder deployment behind a corporate VPN. We do not recommend enabling this option if your deployment is not behind a corporate VPN app."
```

App/Services/CredentialManager.cs
```diff
@@ -223,8 +223,9 @@ private async Task<CredentialModel> LoadCredentialsInner(CancellationToken ct)
             };
         }
 
-        // Grab the lock again so we can update the state.
-        using (await _opLock.LockAsync(ct))
+        // Grab the lock again so we can update the state. Don't use the CT
+        // here as it may have already been canceled.
+        using (await _opLock.LockAsync(TimeSpan.FromSeconds(5), CancellationToken.None))
```
deansheather, line 228: This was a bug I noticed on startup when the server was unreachable, it would timeout the CT and then never set the status to "invalid" so the tray window would always be stuck in the loading state

deansheather (approved):

PR #165, opened 2026-02-23
fix: show sign-in window on launch when not authenticated

### LINKED ISSUE #164: Show sign-in window on first launch when user is not authenticated
**Summary**

When Coder Desktop is launched for the first time (or when the user is not signed in), the app should show the sign-in window instead of silently minimizing to the system tray. Currently, new users see no visible UI after launching the app, which makes it appear broken.

**Problem**

After a fresh install, launching Coder Desktop produces no visible window or taskbar entry. The app goes straight to the system tray as a background process. For a new user who has never used the app before:

1. There is no indication the app launched successfully
2. The user must discover the system tray icon (which Windows often hides behind the ^ overflow arrow)
3. The user must click the tray icon to find the sign-in prompt
4. The app cannot be used at all without signing in, so the silent tray behavior adds unnecessary friction

This leads users to believe the application failed to start — especially since Task Manager shows minimal CPU/memory activity, reinforcing the impression that nothing is happening.

**Proposed Behavior**

• First launch / not signed in: Show the sign-in window automatically. This gives the user a clear next step and confirms the app is running.
• Already authenticated: Continue the current behavior of going silently to the system tray.

Other system tray applications (such as Tailscale, 1Password) all show an onboarding or sign-in window on first launch before settling into tray-only mode.

uzair-coder07:
**This is a Blink Assisted PR. However, I have gone through the changes and tested the resulting executable on a Windows VM in AWS.**

When Coder Desktop launches and the user is not signed in, the app now automatically opens the sign-in window instead of silently minimizing to the system tray. This gives new users a clear next step and confirms the app is running.

Previously, a fresh install produced no visible window or taskbar entry, requiring users to discover the system tray icon to find the sign-in prompt. This led users to believe the application failed to start.

When already authenticated, the existing behavior (going silently to the system tray) is preserved.

**Update:** To avoid being disruptive, the sign-in window is only shown once (on the first launch). Subsequent starts, including computer restarts, preserve the existing silent-to-tray behavior, even if the user still hasn't signed in. Users who are already authenticated are unaffected.

Fixes coder/coder-desktop-windows#164

### DIFF
--- a/App/App.xaml.cs
+++ b/App/App.xaml.cs
@@ -259,20 +259,44 @@ public partial class App : Application, IDispatcherQueueManager, IDefaultNotific
             _logger.LogError($"Failed to refresh sync session state {ex.Message}", ex);
         }
 
-        // If the user is not signed in, automatically show the sign-in window.
-        // This ensures new users see a clear next step instead of the app
-        // silently minimizing to the system tray.
-        var credentialModel = credentialManager.GetCachedCredentials();
-        if (credentialModel.State == CredentialState.Invalid)
+        // If the user is not signed in and we haven't shown the sign-in
+        // window before, show it automatically. This ensures new users see a
+        // clear next step on first launch instead of the app silently
+        // minimizing to the system tray. A marker file is written so the
+        // window is only shown once.
+        try
         {
-            TrayWindow?.DispatcherQueue.TryEnqueue(() =>
+            var credentialModel = credentialManager.GetCachedCredentials();
+            if (credentialModel.State == CredentialState.Invalid && !SignInPromptMarkerExists())
             {
-                var signInWindow = _services.GetRequiredService<SignInWindow>();
-                signInWindow.Activate();
-            });
+                WriteSignInPromptMarker();
+                TrayWindow?.DispatcherQueue.TryEnqueue(() =>
+                {
+                    var signInWindow = _services.GetRequiredService<SignInWindow>();
+                    signInWindow.Activate();
+                });
+            }
         }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Failed to show sign-in window on first launch");
+        }
+
     }
 
+    private static string SignInPromptMarkerPath => Path.Combine(
+        Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+        "CoderDesktop",
+        ".sign-in-prompt-shown");
+
+    private static bool SignInPromptMarkerExists() => File.Exists(SignInPromptMarkerPath);
+
+    private static void WriteSignInPromptMarker()
+    {
+        Directory.CreateDirectory(Path.GetDirectoryName(SignInPromptMarkerPath)!);
+        File.WriteAllBytes(SignInPromptMarkerPath, Array.Empty<byte>());
+    }
+    
     public void OnActivated(object? sender, AppActivationArguments args)
     {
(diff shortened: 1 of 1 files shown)


App/App.xaml.cs
```diff
@@ -258,6 +258,19 @@ private async Task InitializeServicesAsync(CancellationToken cancellationToken =
         {
             _logger.LogError($"Failed to refresh sync session state {ex.Message}", ex);
         }
+
+        // If the user is not signed in, automatically show the sign-in window.
+        // This ensures new users see a clear next step instead of the app
+        // silently minimizing to the system tray.
+        var credentialModel = credentialManager.GetCachedCredentials();
+        if (credentialModel.State == CredentialState.Invalid)
```
rowansmithau: so if the app is signed out, is the logic to show the sign in window? am i reading correctly? if so, i wonder if that might be a bit much, as it will show the sign in window on every start of the computer which would be a pretty drastic UX change for those who don't use coder often, but install the update which this change ships with. I wonder if more of a "show once on first start of app, but not on subsequent starts" approach is more ideal?

deansheather (commented):
I'd argue that instead of activating the sign-in window we should just activate the tray window in the bottom corner of the screen on any **manual** start of the app. The tray window will have a sign-in button if the user isn't signed in.

I think to determine manual vs. automatic launches we'll need to add some flags to the Startup script though.

uzair-coder07:
@deansheather I would say that the tray window would be easy to miss and still requires an extra click. If I recall correctly, Windows collapses overflow tray icons by default, meaning the tray window could easily be hidden behind the "^", right?. The sign-in window, on the other hand, pops up front and center which is hard for any end user to miss out on. 

Also, The tray window's sign-in button would just open the sign-in window anyway, right? So, it seems like we would be adding the flag infrastructure just to add one extra click? At the end of the day, We would not be actually avoiding the sign-in window. We will just be making the user hunt for it is what I am understanding. 

Please let me know if my understanding is incorrect.

App/App.xaml.cs
uzair-coder07: @rowansmithau Hmm, Yup, I have to take this into account. I will add some functionality around this and get back.

uzair-coder07: I have added code to handle this. Also, Tested it out and the pop up is no longer shown when the machine is restarted / shutdown and started up. It is only shown on first install. Any feedback is appreciated!

rowansmithau:
I had a look at this a couple weeks back and think the current implementation since the last commit was added is good, but @uzair-coder07 it might be good to update the original post to summarise the way this implementation works so Dean can give his thoughts with the latest changes?

uzair-coder07:
Thanks Rowan. I will add a comment here and then update the original post.

The following is a summary of how this PR might be beneficial: (cc: @deansheather - whenever you get a chance)

When Coder Desktop launches and the user is not signed in, the app now automatically opens the sign-in window front and center, giving new users a clear next step immediately after install.

Previously, the app would silently minimize to the system tray with no visible window or taskbar entry, leaving users to discover the tray icon on their own. This led users to believe the app failed to start.

To avoid being disruptive, the sign-in window is only shown once (on the first launch). Subsequent starts, including computer restarts, preserve the existing silent-to-tray behavior, even if the user still hasn't signed in. Users who are already authenticated are unaffected.

deansheather (approved):

PR #175, opened 2026-07-13
feat: send WakeRequest to the tunnel on system resume

### LINKED ISSUE #172: Send CoderVPN WakeRequest on system wake to recover tunnel paths
## Summary

After the machine sleeps and wakes, the Coder Connect tunnel can stay unusable for several minutes: new SSH dials fail with `i/o timeout` until magicsock's periodic re-STUN re-homes the path on its own. A short, same-network wake often produces no netmon link-change event, so the daemon's existing `linkChange()` recovery (`Rebind` + `ReSTUN`) never fires.

The same gap was first observed on macOS (coder/coder#26736), but it is OS-agnostic: the daemon-side fallback timer also stops after 45s idle, and sleep is idle by definition, so on wake there is no scheduled re-STUN until another trigger fires.

## Background

coder/coder PR [https://github.com/coder/coder/pull/26739](<https://github.com/coder/coder/pull/26739>) adds the daemon-side hook. It introduces a new CoderVPN RPC (`WakeRequest` / `WakeResponse`, tunnel protocol `1.3`) whose handler calls `tailnet.Conn.Rebind()` (`magicConn.Rebind()` + `magicConn.ReSTUN("wake")`), the same recovery path a major link change triggers.

The daemon side only exposes the hook. Coder Desktop must actually emit it. This is the Windows counterpart of coder/coder-desktop-macos#260.

## What's needed in coder-desktop-windows

1. Subscribe to a system resume event. In .NET this is `Microsoft.Win32.SystemEvents.PowerModeChanged` with `PowerModes.Resume`; for a service/headless context without a message pump, register for `WM_POWERBROADCAST` (`PBT_APMRESUMESUSPEND` / `PBT_APMRESUMEAUTOMATIC`) via `RegisterPowerSettingNotification`.
2. On resume, send a `ManagerMessage` with the new `WakeRequest` over the existing CoderVPN tunnel RPC channel.
3. Only send when the negotiated tunnel protocol version is `>= 1.3` (older daemons won't handle the message). Regenerate the C# protobuf bindings from the updated `vpn.proto` once the coder/coder PR merges.
4. Optionally surface `WakeResponse.success` / `error_message` in debug logs.

This is an event-driven kick on a known OS event, not a health check or poll. It forces path re-discovery and never tears down a live tunnel.

## Refs

* Root cause and analysis: [https://github.com/coder/coder/issues/26736](<https://github.com/coder/coder/issues/26736>)
* Daemon-side implementation: [https://github.com/coder/coder/pull/26739](<https://github.com/coder/coder/pull/26739>)
* macOS counterpart: [https://github.com/coder/coder-desktop-macos/issues/260](<https://github.com/coder/coder-desktop-macos/issues/260>)

EhabY:
Fixes coder/coder-desktop-windows#172

After the machine sleeps and wakes, the Coder Connect tunnel can stay unusable for several minutes until magicsock's periodic re-STUN recovers the path on its own. The daemon side (coder/coder#26739, tunnel protocol `1.3`) added a `WakeRequest` RPC that forces network path re-discovery on demand. This PR makes the Windows manager emit it on system resume.

## Changes

* `Vpn.Proto/vpn.proto`: synced with upstream `coder/coder`; adds `WakeRequest` / `WakeResponse` (bindings generated at build time).
* `RpcVersion`: `Current` bumped `1.1` -> `1.3`; new `IsAtLeast()` ordered comparison for version-gating.
* `Speaker`: exposes `NegotiatedVersion`, captured during the handshake (previously discarded).
* `TunnelSupervisor`: surfaces the tunnel's `NegotiatedVersion` through `ITunnelSupervisor`.
* `SystemResumeMonitor` (new): event-based class raising `Resumed` from `SystemEvents.PowerModeChanged`; registered as a singleton.
* `Manager`: subscribes to `Resumed` (mirroring the existing `IManagerRpc.OnReceive` wiring) and sends `WakeRequest` when the negotiated version is `>= 1.3`, with a 5s reply timeout. The send is fire-and-forget off the SystemEvents broadcast thread; `SendWakeRequest` logs and swallows all failures by construction, so nothing can fault the discarded task or affect the running tunnel.

## Tests

* `SpeakerTest`: negotiated version exposed after handshake.
* `RpcVersionTest`: `IsAtLeast` ordering.
* `ManagerTest` (new): wake request sent on the resume event; skipped when the tunnel isn't running or the version is too old; send failures are swallowed.

<details><summary>Implementation decisions</summary>

1. `1.2`'s `last_ping` fields were already present in this repo's proto; only the version constant was never bumped, so `Current` goes straight to `1.3`.
2. Resume detection uses `SystemEvents.PowerModeChanged` (as suggested in the issue): works in both service and console mode, and SystemEvents runs its own broadcast window thread so no message pump is needed. The service-control-handler alternative (`CanHandlePowerEvent`/`OnPowerEvent`) requires subclassing `WindowsServiceLifetime` and doesn't fire in console/debug mode.
3. Version gating uses standard ordering (`IsAtLeast`) rather than an exact-major match, so a future major protocol bump keeps sending wake requests.
4. Wake is a unary RPC with a 5s timeout rather than pure fire-and-forget at the RPC layer, so unexpected replies can be logged.
5. The wake gate relies on `NegotiatedVersion == null` to detect a stopped tunnel, keeping `SendWakeRequest` side-effect free and unit-testable with a fake `ITunnelSupervisor`.

</details>

---

*This PR was generated with Coder Agents on behalf of* @EhabY\*.\*

### DIFF
--- a/Tests.Vpn.Proto/RpcVersionTest.cs
+++ b/Tests.Vpn.Proto/RpcVersionTest.cs
@@ -41,14 +41,14 @@ public class RpcVersionTest
         IsCompatibleWithBothWays(twoOne, new RpcVersion(1, 1), Is.Null);
     }
 
-    [Test(Description = "Check feature support against the version that introduced the feature")]
-    public void SupportsFeature()
+    [Test(Description = "Compare versions with IsAtLeast")]
+    public void IsAtLeast()
     {
-        var featureVersion = new RpcVersion(1, 3);
-        Assert.That(new RpcVersion(1, 3).SupportsFeature(featureVersion), Is.True);
-        Assert.That(new RpcVersion(1, 4).SupportsFeature(featureVersion), Is.True);
-        Assert.That(new RpcVersion(1, 2).SupportsFeature(featureVersion), Is.False);
-        Assert.That(new RpcVersion(2, 3).SupportsFeature(featureVersion), Is.False);
+        var oneThree = new RpcVersion(1, 3);
+        Assert.That(oneThree.IsAtLeast(oneThree), Is.True);
+        Assert.That(new RpcVersion(1, 4).IsAtLeast(oneThree), Is.True);
+        Assert.That(new RpcVersion(2, 0).IsAtLeast(oneThree), Is.True);
+        Assert.That(new RpcVersion(1, 2).IsAtLeast(oneThree), Is.False);
     }
 }
 
--- a/Tests.Vpn.Proto/RpcVersionTest.cs
+++ b/Tests.Vpn.Proto/RpcVersionTest.cs
@@ -40,6 +40,18 @@ public class RpcVersionTest
         // 2.1 && 1.1 => null
         IsCompatibleWithBothWays(twoOne, new RpcVersion(1, 1), Is.Null);
     }
+
+    [Test(Description = "Check feature support against the version that introduced the feature")]
+    public void SupportsFeature()
+    {
+        var featureVersion = new RpcVersion(1, 3);
+        Assert.That(new RpcVersion(1, 3).SupportsFeature(featureVersion), Is.True);
+        Assert.That(new RpcVersion(1, 4).SupportsFeature(featureVersion), Is.True);
+        Assert.That(new RpcVersion(1, 2).SupportsFeature(featureVersion), Is.False);
+        // Major versions must match exactly.
+        Assert.That(new RpcVersion(2, 3).SupportsFeature(featureVersion), Is.False);
+        Assert.That(new RpcVersion(2, 4).SupportsFeature(featureVersion), Is.False);
+    }
 }
 
 [TestFixture]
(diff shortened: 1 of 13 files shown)


EhabY:
I've tested this on Windows 11 and it does produce this event even if it's a short sleep or a long one (several hours)

ibetitsmike:
Tested the flow and it seems to improve the reconnection speed. Code LGTM

ibetitsmike (approved):

matifali:
We would also need to backport the upstream fix once verified to the stable and ESR channel.

spikecurtis:
@matifali I don't believe such channels exist for Coder Desktop Windows

matifali:
I mean the upstream PR this depends on: coder/coder#26739<br>I may be mistaken, but my understanding is that this needs that PR in the Coder server.

EhabY:
Yes it needs that PR but this event is only emitted if the protocol matches (1.3), so if the server is old then this is a no-op

PR #180, opened 2026-09-04
fix: send coder User-Agent on all deployment requests

### LINKED ISSUE #176: Desktop sends no User-Agent header - all API requests blocked by AWS WAF (NoUserAgent_HEADER) on self-hosted deployments
## Summary

Coder Desktop for Windows sends HTTP requests with **no `User-Agent` header at all**. Any deployment sitting behind AWS WAF with the `AWSManagedRulesCommonRuleSet` managed rule group will reject every one of those requests with `403 Forbidden` - the `NoUserAgent_HEADER` rule is part of that rule set and blocks by default.

Result: sign-in fails on `GET /api/v2/buildinfo` and the app is completely unusable against the deployment. The Coder CLI, which does set a User-Agent, works normally against the exact same deployment from the same machine.

## Environment

- Coder Desktop: **0.8.4** (also 0.8.4.0 for "Coder Desktop (Core)")
- Coder CLI: 2.28.4 (works fine)
- OS: Windows 11 Pro 24H2 (10.0.26100)
- Deployment: self-hosted Coder on AWS, behind an Application Load Balancer with an AWS WAF Web ACL using `AWSManagedRulesCommonRuleSet`

## Reproduction

Any public, unauthenticated Coder endpoint behind such an ALB shows it. With a User-Agent:

```
$ curl -s -D - -o /dev/null https://<deployment>/api/v2/buildinfo | head -6
HTTP/2 200
date: ...
content-type: application/json; charset=utf-8
content-length: 476
set-cookie: csrf_token=...; Path=/; HttpOnly; Secure; SameSite=Lax
vary: Origin
```

With the User-Agent suppressed (`-H 'User-Agent;'`), which is what Desktop does:

```
$ curl -s -D - -o /dev/null -H 'User-Agent;' https://<deployment>/api/v2/buildinfo | head -6
HTTP/2 403
server: awselb/2.0
date: ...
content-type: text/html
content-length: 118
```

The 200 response carries Coder application headers (`x-coder-build-version`, `x-coder-request-id`, `x-ratelimit-*`), so it reached the app. The 403 carries only `server: awselb/2.0` - blocked at the edge, never reaching Coder.

## Expected behaviour

Coder Desktop should set a User-Agent on every outbound HTTP request, e.g. `Coder-Desktop/0.8.4 (windows; x64)`, consistent with what the CLI already does.

## Impact

This is not specific to one deployment. `AWSManagedRulesCommonRuleSet` is the default recommendation in AWS's own WAF documentation and is very widely deployed, so any self-hosted Coder behind AWS WAF with default managed rules will be unreachable from Coder Desktop. The only workaround today is asking the platform team to add a WAF exception allowing requests with a missing/empty User-Agent — which is a security regression they are reasonably reluctant to make.

## Suggested fix

Set a default `User-Agent` on the shared `HttpClient` used by the Desktop app (and the Core/service component) rather than relying on the framework default, which omits the header entirely.

untra:
* [DEVEX-778](https://linear.app/codercom/issue/DEVEX-778/desktop-sends-no-user-agent-header-all-api-requests-blocked-by-aws-waf)
## Summary
.NET's HttpClient omits `User-Agent` by default, causing AWS WAF to reject requests with `403` on deployments behind an ALB. This blocked sign-in (`/api/v2/buildinfo`, `/api/v2/users/me`), and — because Desktop downloads the tunnel binary from the deployment rather than bundling it — Coder Connect could not start either. Closes #176.

## What changed
- **`CoderSdk/UserAgent.cs`** (new): formats `User-Agent` as `<token>/<version> (<goos>/<goarch>)`,
  e.g. `coder-desktop/0.8.4 (windows/amd64)`. Platform names follow Go's `GOOS`/`GOARCH` rather than
  .NET's spelling, so a single pattern matches the CLI, the vpn-daemon and every Desktop client.

- **SDK HTTP clients**: `JsonHttpClient` — the single chokepoint both `CoderApiClient` and
  `AgentApiClient` route through — now sets `DefaultRequestHeaders.UserAgent`, selected by a
  `CoderComponent` enum (`Desktop` / `Core`). `DefaultRequestHeaders` is deliberate: .NET applies it
  only when the request doesn't already carry the header, so caller-supplied headers still win.
- **`Vpn.Service`**: `Manager.cs` and `Downloader.cs` identify as `CoderComponent.Core`, covering the
  buildinfo/users-me calls and the tunnel binary download.

## Validation
- `Tests.CoderSdk`: **8 → 14 passing** (+6 `UserAgentTest`). 
- `Tests.Vpn.Service/DownloaderTest.cs`: **26 → 28** (+`WithUserAgent`, +`WithUserAg
  Both assert the header on the wire via the existing `TestHttpServer` fixture, including that a
  caller-supplied `User-Agent` overrides the default.
- **Tested build in parallels**: `dotnet build` clean, built and ran the arm build with changes, run in windows parallels
Screenshot 2026-09-04 at 7 23 01 AM

### DIFF
--- a/Tests.CoderSdk/UserAgentTest.cs
+++ b/Tests.CoderSdk/UserAgentTest.cs
@@ -1,0 +1,55 @@
+using System.Reflection;
+using System.Text.RegularExpressions;
+using Coder.Desktop.CoderSdk;
+
+namespace Coder.Desktop.Tests.CoderSdk;
+
+[TestFixture]
+public class UserAgentTest
+{
+    // Every Coder client is expected to match this, so operators can allowlist the token set. Keep it in
+    // sync with the Coder CLI (cli/root.go) and the macOS and Linux Desktop clients.
+    private static readonly Regex Grammar = new(
+        @"^coder-(cli|desktop|desktop-core|vpn-daemon)/\d+\.\d+\.\d+ \((windows|darwin|linux|unknown)/(386|amd64|arm|arm64|unknown)(; .+)?\)$");
+
+    [Test(Description = "Matches the shared User-Agent grammar")]
+    [TestCase(CoderComponent.Desktop)]
+    [TestCase(CoderComponent.Core)]
+    public void MatchesGrammar(CoderComponent component)
+    {
+        Assert.That(UserAgent.Build(component), Does.Match(Grammar));
+    }
+
+    [Test(Description = "Each component reports its own token")]
+    public void ComponentTokens()
+    {
+        Assert.That(UserAgent.Build(CoderComponent.Desktop), Does.StartWith("coder-desktop/"));
+        Assert.That(UserAgent.Build(CoderComponent.Core), Does.StartWith("coder-desktop-core/"));
+    }
+
+    [Test(Description = "Four-part assembly versions are trimmed to the three-part release version")]
+    public void TrimsAssemblyVersion()
+    {
+        var assembly = Assembly.GetExecutingAssembly();
+        var version = assembly.GetName().Version;
+        Assert.That(version, Is.Not.Null, "test assembly has no version");
+
+        var ua = UserAgent.Build(CoderComponent.Desktop, assembly);
+        Assert.That(ua, Does.Match(Grammar));
+        Assert.That(ua, Does.StartWith($"coder-desktop/{version!.Major}.{version.Minor}.{version.Build} ("));
+    }
+
+    [Test(Description = "An unversioned assembly still produces a well-formed User-Agent")]
+    public void UnknownVersion()
+    {
+        var ua = UserAgent.Build(CoderComponent.Desktop, null);
+        Assert.That(ua, Does.Match(Grammar));
+        Assert.That(ua, Does.StartWith("coder-desktop/0.0.0 ("));
+    }
+
+    [Test(Description = "Unrecognized components are rejected rather than silently mislabeled")]
+    public void UnknownComponent()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => UserAgent.Build((CoderComponent)(-1)));
+    }
+}
--- a/Vpn.Service/Downloader.cs
+++ b/Vpn.Service/Downloader.cs
@@ -6,6 +6,7 @@ using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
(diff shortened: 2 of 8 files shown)


Tests.CoderSdk/UserAgentTest.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Coder.Desktop.Tests.CoderSdk;
+
+[TestFixture]
+public class UserAgentTest
+{
+    // Every Coder client is expected to match this, so operators can allowlist the token set. Keep it in
+    // sync with the Coder CLI (cli/root.go) and the macOS and Linux Desktop clients.
+    private static readonly Regex Grammar = new(
+        @"^coder-(cli|desktop|desktop-core|vpn-daemon)/\d+\.\d+\.\d+ \((windows|darwin|linux|unknown)/(386|amd64|arm|arm64|unknown)(; .+)?\)$");
```
jeremyruppel, line 13: question: is this whole regex necessary? looks like we only have `coder-desktop` and `coder-desktop-core` UA prefixes. not that this is incorrect, in fact I love the parity, but I'm wondering if the regex complexity and drift potential is worth it here

Vpn.Service/Downloader.cs
```diff
@@ -348,10 +349,16 @@ public class DownloadTask
...
     {
-        AutomaticDecompression = DecompressionMethods.All,
-    });
+        var client = new HttpClient(new HttpClientHandler
+        {
+            AutomaticDecompression = DecompressionMethods.All,
+        });
+        client.DefaultRequestHeaders.UserAgent.ParseAdd(UserAgent.Build(CoderComponent.Core));
+        return client;
+    })();
```
jeremyruppel, line 360: I'm not a .NET wiz. is this iife idiomatic .NET?

untra: I asked gemini for an optimization of a sloppier opus implementation and got this. Def sagreed I haven't written .NET like this before, but creating and calling an anonymous function to also do the 

`client.DefaultRequestHeaders.UserAgent.ParseAdd(UserAgent.Build(CoderComponent.Core));`

makes this a lot cleaner.

Tests.CoderSdk/UserAgentTest.cs
jeremyruppel: per discussion in slack, this is probably best to keep in sync

jeremyruppel (approved):

